PR #7, opened 2020-04-06
Normalization of spectra

jsitarek:
A few changes concerning normalization of spectra, including a fix for U_e normalization and new funcitonality to normalize at gamma=1


cosimoNigro:
Hey @jsitarek,

thanks. To use this new normalisation you introduced it appears we have to do something like this
```python
# set the spectrum normalisation (total energy in electrons in this case) 
spectrum_norm = 0 
# define the spectral function through a dictionary 
spectrum_dict = { 
    "type": "PowerLaw", 
    "parameters": {"k_1":1e-13 * u.Unit("cm-3"), "p": 2.8, "gamma_min": 1, "gamma_max": 1e7} 
} 
R_b = 1e16 * u.cm 
B = 1 * u.G 
z = Distance(1e27, unit=u.cm).z 
delta_D = 10     
Gamma = 10 
blob = Blob(R_b, z, delta_D, Gamma, B, spectrum_norm, spectrum_dict)                                                                                                     
normalising power-law to value at gamma=1: 1.00e-13 1 / cm3
```
The implementation is ok, the only thing I do not like the idea that the norm has to be zero and that the an additional parameter (`k_1`) has to be introduced in the spectrum dict. 
I would set a flag in the class to distinguish which kind of normalisation we are using. 
Hope is ok if I merge and continue on your work.

jsitarek:
sure, I had to use a different syntax that what you were using for the normalization to total particle density and to energy density because cm-3 as a unit would be ambiguous in the code. I added the new parameter to make it at least similar to the other parameters, but sure making a flag is another possibility.

Issue #6, opened 2020-04-06
SmoothlyBrokenPowerLaw normalization


jsitarek:
SBPL normalization from energy density does not work (it gives an error because of wrong units).

BTW, according to the definition in the code, SmoothlyBrokenPowerLaw is not "smoothly" broken, there is a sharp break at gamma_b

cosimoNigro:
Thanks,

was this solved by PR #8?

jsitarek:
Hi, I confirm that the problem was solved, and also the name of the spectral deifnition now looks more consistent. Still I'm wondering if it is useful to keep both BrokenPowerLaw and BrokenPowerLaw2. They only differ by definition of k_e and except of this when e.g. initialized with normalization to (energy) density they will provide exactly the same results. The only different is k_e, but if you keep the original BrokenPowerLaw, one can also normalize it to k_e at gamma=1 with those new functions.

PR #8, opened 2020-04-06
Improve handling of the spectrum normalisation

cosimoNigro:
Dear @jsitarek,

this PR is a follow-up to PR #7.
I introduced a new parameter in the `Blob` initialiser, namely `spectrum_norm_type` to allow for three kinds of normalisation: `["integral", "differential", "gamma=1"]`. `"integral"` is the default, it sets the spectrum such that the value in `spectrum_norm` is its integrand. The other two are available only for `cm-3` units. `"differential"`sets the `k_e` parameter directly to the value in `spectrum_norm` while `"gamma=1"` makes use of the functions you implemented in PR #7, i.e. `n_e(gamma=1)` is set to the value in `spectrum_norm`.

I updated the documentation and added an example, try from the main folder
```python
python tests/basic/test_spectra.py
```
(the macro is supposed to trigger a final error).

I think I also fixed Issue #6.


cosimoNigro:
https://agnpy.readthedocs.io/en/latest/emission_regions.html#normalisation-modes
Let me know if the usage is clear.

PR #9, opened 2020-04-17
added tutorial notebook fitting SSC with gammapy, fit not converging yet

cosimoNigro:
With this PR I have added a notebook to attempt to fit a Mrk421 SED with gammapy's methods.
I tried to cast the full SED into a `SpectralModel` and fit a series of `FluxPoints`.
The parameters left free in the fit are the electron distribution spectral parameters and the magnetic field.
The fit is not converging yet, I am merging though since I need to present it to the gammapy developer's call.


cosimoNigro:
I have done some minor modification in the `emission_regions` and in `spectra`, the high-level interface is the same though.

PR #15, opened 2020-05-10
break and maximum energy from SSC losses 

jsitarek:
functions to compute magnetic field and synchrotron photons energy densities
functions using synchr. photon density to calculate the corresponding energy break and max energy


cosimoNigro:
Thanks @jsitarek, I am merging.

PR #21, opened 2020-05-15
functions for sed peaks

cosimoNigro:
This small pull request adds functions to compute the peak flux and frequencies of the SEDs.

@jsitarek watchout I changed the names of two functions you introduced:
`u_B` -> `U_B` (in Finke and Dermer is always capitalised though other energy densities are not)
`u_dens_synchr` -> `u_ph_synch` (I have named the other functions computing the densities of photons as `u_ph`, also in the whole code we use _synch rather than _synchr).

Sorry for not asking to change that in the PR, I did not check carefully.
I modified test_gmax.py to work with the new functions, but as I said I'll make proper test scripts soon.

### DIFF
--- a/agnpy/compton.py
+++ b/agnpy/compton.py
@@ -237,6 +237,17 @@ class SynchrotronSelfCompton:
         sed = prefactor * self.com_sed_emissivity(epsilon_prime)
         return sed.to("erg cm-2 s-1")
 
+    def sed_peak_flux(nu):
+        """provided a grid of frequencies nu, returns the peak flux of the SED
+        """
+        return self.sed_flux(nu).max()
+
+    def sed_peak_nu(nu):
+        """provided a grid of frequencies nu, returns the frequency at which the SED peaks
+        """
+        idx_max = self.sed_flux(nu).argmax()
+        return nu[idx_max] 
+
 
 class ExternalCompton:
     """class for External Compton radiation computation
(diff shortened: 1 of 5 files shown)


jsitarek:
u_B is used also in P_jet_B (sorry it is my fault, you were using before just B^2 / 8 pi, but since I also needed it for another function when I introduced the function I changed yours as well to simplify it and forgot to tell you.)
        return (prefactor * self.u_B).to("erg s-1")
https://github.com/cosimoNigro/agnpy/blob/738a445b53f1daebbeaf598d32de389249e077e8/agnpy/emission_regions.py#L282

cosimoNigro:
Fixed in master, thank you.

jsitarek:
I just realized that u_ph_synch is somewhat misleading name. You use u_ph in other places of the code, but it is for the photon density, and this function is energy density

cosimoNigro:
No, it is also density of radiation https://github.com/cosimoNigro/agnpy/blob/master/agnpy/targets.py#L192
see corresponding formulas in Finke's paper.

cosimoNigro:
Also check the classical Tavecchio's plot for the densities (Fig 14. of
https://www.aanda.org/articles/aa/pdf/2014/09/aa23484-14.pdf) 
it's always energy density, not simple volumetric density. 
Right?

jsitarek:
right, sorry, my mistake, should have seen that the returned units are erg cm-3, but then the naming is not very fortunate, since 'ph' normally stands for photons, so photon density is number density of photons, not eneryg density. 
Actually Finke uses also n_ph but in sense of number density, not energy density (see beginning of section 2.1):
"${n}_{{ph}}(\epsilon ,{\rm{\Omega }})$ is the number density of incident ("seed") photons with energies between $\epsilon $ and epsilon + depsilon and solid angles between Ω and ${\rm{\Omega }}+d{\rm{\Omega }}$,"

PR #28, opened 2020-05-26
misleading message and updated EC tests

jsitarek:
fixed misleading message in setting normalization for broken PWL
updated tests to the new API of EC


cosimoNigro:
I'll take care of solving the conflicts tomorrow!

PR #30, opened 2020-06-07
CMB and point like sources added to targets; energy densities and EC implemented for all targets

cosimoNigro:
This PR add some features in view of release 0.0.6:

- added CMB and point like source behind the jet as photon targets in `targets.py`;

- added energy densities computation in the stationary and coming frame (`u` and `u'`) for all targets;

- added a notebook tutorial on energy densities;

- added EC on CMB and point like source in `compton.py`;

- added also a script (`experiments/basic/ec_point_like_comparison.py`)
   to check that in the limit of large distances the SED for EC on BLR and dust torus
   reduce to the SED for EC on point sources approximating them;

- add to the docs all calculations implemented for the energy densities and the SEDs.


cosimoNigro:
The failing tests are related to the code coverage, new classes and functions have been added without the corresponding tests, will do in a follow-up PR.

PR #35, opened 2020-07-20
Include copy of license in the source distribution

mwcraig:
This pull request adds the license file to the source distribution. Having it there makes it a bit easier to package agnpy for conda.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ The code is available in the [python package index](https://pypi.org/project/agn
 pip install agnpy
 ```
 
+The code can also be installed using `conda`
+
+```bash
+conda install -c conda-forge agnpy
+```
+
 ## tests
 A test suite is available in [`agnpy/tests`](https://github.com/cosimoNigro/agnpy/tree/master/agnpy/tests), to run it just type
 `pytest` in the main directory.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -14,7 +14,7 @@ Notation and basic formulas are borrowed from [DermerMenon2009]_ which constitut
 
 Implementation
 ..............
-The numerical operations are delegated to `numpy arrays <https://numpy.org>`_, all the physical quantities computed are casted as `astropy Quantities <https://docs.astropy.org/en/stable/units/>`_. 
+The numerical operations are delegated to `numpy arrays <https://numpy.org>`_, all the physical quantities computed are casted as `astropy Quantities <https://docs.astropy.org/en/stable/units/>`_.
 
 License
 -------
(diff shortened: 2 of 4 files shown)


cosimoNigro:
Hi @mwcraig, thanks for adding the `conda-recipe` for `agnpy`.
You listed yourself as `recipe-maintainer` in the `conda-recipe`, will you take care of updating the recipe when I release a new version?

mwcraig:
Hi @cosimoNigro! 

>You listed yourself as recipe-maintainer in the conda-recipe, can I assume that you will take care - in the future - of updating the recipe when I release a new version?

I will do as soon as the package shows up on conda-forge -- right now it looks like it takes a while from merging the recipe to creating the new github repo. We don't  need to wait for a new release to add you (unless you prefer to do that). Getting the initial recipe added tends to go faster with fewer maintainers.

> Could you also add to this PR a commit updating the README.md and the installation section in the docs specifying that now you can install agnpy with conda?

I'm happy to, it is on the way! Note that at the moment conda install still won't work, so you may want to wait until it works to merge this.

README.md
```diff
@@ -29,6 +29,12 @@ The code is available in the [python package index](https://pypi.org/project/agn
 pip install agnpy
 ```
 
+The code can also be installed using `conda`
+
+```bash
+conda install -c conda-forge agnpy
```
mwcraig, line 35: If you prefer, could use the `astropy` channel here instead. It is the same underlying package in either case.

docs/index.rst
```diff
@@ -14,20 +14,26 @@ Notation and basic formulas are borrowed from [DermerMenon2009]_ which constitut
...
 
 .. code-block:: bash
 
     pip install agnpy
 
+The code can also be installed with ``conda``
+
+.. code-block:: bash
+
+    conda install -c conda-forge agnpy
```
mwcraig, line 35: Again, if you prefer, the channel can be changed to `astropy`.

README.md
cosimoNigro: It's fine as is

docs/index.rst
cosimoNigro: It's fine as is

cosimoNigro:
> I will do as soon as the package shows up on conda-forge -- right now it looks like it takes a while from merging the recipe to creating the new github repo. We don't need to wait for a new release to add you (unless you prefer to do that). Getting the initial recipe added tends to go faster with fewer maintainers.

I am fine with not being added to the maintainers

> I'm happy to, it is on the way! Note that at the moment conda install still won't work, so you may want to wait until it works to merge this.

Ok just give me a heads up when the package shows up on conda-forge and I'll merge this PR.

Thanks again.

mwcraig:
The package is on conda-forge and on the astropy channel, so merge at will!

Issue #38, opened 2020-08-04
absorption example error


pawel21:
Hi,

I have tried use absorption example [doc](https://agnpy.readthedocs.io/en/latest/absorption.html)
and I got the following error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-b07779d8d11b> in <module>
      2 r = 1.1e16 * u.cm
      3 
----> 4 absorption_disk = Absorption(blob, disk, r=r)
      5 absorption_blr = Absorption(blob, blr, r=r)
      6 absorption_dt = Absorption(blob, dt, r=r)

~/anaconda3/envs/agn/lib/python3.7/site-packages/agnpy-0.0.7-py3.7.egg/agnpy/absorption.py in __init__(self, blob, target, r)
     43         self.target = target
     44         self.r = r
---> 45         self.set_mu()
     46         self.set_phi()
     47         self.set_l()

~/anaconda3/envs/agn/lib/python3.7/site-packages/agnpy-0.0.7-py3.7.egg/agnpy/absorption.py in set_mu(self, mu_size)
     49     def set_mu(self, mu_size=100):
     50         self.mu_size = mu_size
---> 51         if self.target.type == "SSDisk":
     52             # in case of hte disk the mu interval does not go from -1 to 1
     53             r_tilde = (self.r / self.target.R_g).to_value("")

AttributeError: 'SSDisk' object has no attribute 'type'
```
I guess this is due to disk not have attribute 'type', but 'name'.
Could you reploduce my error?

cosimoNigro:
Hi @pawel21,

thanks for reporting. You are exactly right, I changed the targets but forgot to update the `Absorption` module.
I planned to do it in a second moment since I had some issues to fix related to that module (e.g. Issue #16).
Will take a look this week.

Cheers!

cosimoNigro:
You are also very welcome to fix it yourself, if you feel you have time.
The fix on #16 requires a bit of thought but the one is preventing you to run the tutorial does not ;)

pawel21:
Thank you, I will fix it.

cosimoNigro:
Solved by PR #39 
thanks @pawel21 !

PR #39, opened 2020-08-04
Absorbtion name fix 

pawel21:
Issue #38
I have fixed name in `Absorption` module

### DIFF
--- a/agnpy/absorption.py
+++ b/agnpy/absorption.py
@@ -2,6 +2,7 @@ import numpy as np
 from astropy.constants import h, c, m_e, sigma_T
 import astropy.units as u
 from .compton import cos_psi, x_re_shell, x_re_ring, mu_star
+from .targets import SSDisk, SphericalShellBLR, RingDustTorus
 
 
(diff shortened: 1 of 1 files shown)


cosimoNigro:
Thanks @pawel21, would you mind doing the selection using `isinstance` (instead of `.name == "somestring"`); see an example in the external compton SED computation:
https://github.com/cosimoNigro/agnpy/blob/master/agnpy/compton.py#L556
This way even if we add / remove an attribute (or change type to name as I did) the code will not be affected as it will search for an instance of a given class.

pawel21:
> Thanks @pawel21, would you mind doing the selection using `isinstance` (instead of `.name == "somestring"`); see an example in the external compton SED computation:
> https://github.com/cosimoNigro/agnpy/blob/master/agnpy/compton.py#L556
> This way even if we add / remove an attribute (or change type to name as I did) the code will not be affected as it will search for an instance of a given class.

Yes, sure. This is better way.

pawel21:
I changed to use `isinstance`

PR #40, opened 2020-08-06
 wrong function name for photon energy density and error in printing of broken power law

jsitarek:
- fixed wrong function name for photon energy density
- fixed printing of broken power law


cosimoNigro:
Thanks!
I think we should move all these functions constraining the parameters in another class, what do you think?

Issue #16, opened 2020-05-14
issue with mu, cosine of zenith, in absorption and radiation calculation from disk


cosimoNigro:
The range of mu used for integration changes with the position of the blob above the BH (r) according to

```
 mu_min = 1 / sqrt(1 + (R_out / r)^2)
 mu_max = 1 / sqrt(1 + (R_in / r)^2)
```

In the absorption though an array of mu calculated at a single r (the one occupied by the blob) is used
https://github.com/cosimoNigro/agnpy/blob/master/agnpy/absorption.py#L97
when integrating from r to infinity for the optical depth, the array of mu has to be recomputed at each value of r and updated.

For the same reason the same problem might occur in the photon density:
https://github.com/cosimoNigro/agnpy/blob/master/agnpy/targets.py#L202

The integration has to be separated somehow, I'll take care of that.

cosimoNigro:
The integration is now properly performed by changing the extreme of integrations in `mu` at each distance.
Taken care by PR #46

PR #46, opened 2020-10-19
Improving the optical depth calculation

cosimoNigro:
Hello,

This PR tries to improve the optical depth calculations.
It solves issue #16 though the agreement with Finke is not yet good.

Here a comparison of the optical depths for Disk
tau_disk_comaprison_figure_14_finke_2016
agnpy absorption starts much earlier (a factor ten in energy) than the one in Finke (2016). 
Additionally agnpy absorption uses the full integration: if I try to use the simplified formula in Finke (2016) - green line (blob moving parallel to the jet) I get an absorption similarly starting at lower energies than the one of the figure sampled from the paper plot (blue line).

The absorption for the Broad Line Region and Dust Torus instead starts at larger energies than in Finke
tau_dt_comparison_figure_14_finke_2016
tau_blr_lyman_alpha_comparison_figure_14_finke_2016

Note though that agnpy's absorption is self-consistent. If I replace the dust torus and the BLR with a monochromatic point-like source approximating them at large distances (larger than the size of the target) I have a very good agreement (I did the same tests for the external Compton).
tau_dt_point_source_comparison
tau_blr_point_source_comparison

I don't know what am I doing wrong, any suggestion @jsitarek, @pawel21?
@jsitarek you still have that code you used to compute optical depths, might we use it for crosschecks?


cosimoNigro:
I am merging, let us continue the discussion with issues.

PR #52, opened 2020-11-16
General restructuring of the SED for their usage as fittable models

cosimoNigro:
This PR implements many internal changes aimed at creating functions suited to be passed to a fitting procedure.

1. I introduced `staticmethod`s for each of the radiative processes such that the SED can be computed changing all the parameters together.
The problem was this: if we wanted to fit an external Compton SED with the previous version of the code,
half of the parameters were contained in the `Blob` instance, the other half in the `target` instance. 
Therefore a fitting procedure would have needed to access the objects individually, change them, and redefine the `ExternalCompton` class with them.
Now everything can be done within a single function containing all the parameters of the model, e.g.
```python
Synchrotron.evaluate_sed_flux(
    nu, z, d_L, delta_D, B, R_b, LogParabola, k_e, p, q, gamma_0, gamma_min, gamma_max
)
```
for the Synchrotron radiation
```python
SynchrotronSelfCompton.evaluate_sed_flux(
    nu, z, d_L, delta_D, B, R_b, LogParabola, k_e, p, q, gamma_0, gamma_min, gamma_max
)
``` 
for the SSC.
You can find in `experiments/basic/sherpa_fitting_ssc_trial.ipynb` an example of how I plugged these new functions in `sherpa` to create a fittable model. The fit is not decent but I did not play that much with the parameters! 
I just wanted to be sure that the `agnpy` model could be coupled to `sherpa`.

Of course the high-level function `sed_flux(nu)` has remained the same.

2. I introduced a new integration procedure half-copied from `naima`. 
It is an implementation of the trapezoidal rule in log-space. Each slice of the function is interpolated with a power-law rather than with a straight line. It is slightly more precise of the trapezoidal rule in linear space. Both for the spectra and for the radiative processes one can choose whether to use the trapezoidal rule in log-log space or the normal one when integrating through the `integrator` argument.
Here a comparison of the SED for EC on Disk computed with both the integration methods.
ec_disk_comparison_integration_methods

3. I added some functionalities in `utils` that we often use (conversion, geometric computation, plotting routines).

4. The **only major change in the API** is this: **it is not needed anymore to pass the Synchrotron object to the SynchrotronSelfCompton class.** It was just redundant.
`SynchrotronSelfCompton(blob, synchro)` is now just `SynchrotronSelfCompton(blob)` 
I updated the docs accordingly.

I marked this modifications as 0.0.7.9, If anyone wants to try it, please do it.
We can make the 0.0.8 release in few days.


cosimoNigro:
@jsitarek I apologise if this breaks the changes you were doing to check the absorption.
Notice now I created a submodule for each component (emitters and radiative processes).
I thought I could make this changes faster but then the preparation for the MAGIC meeting and the ADASS got me stuck.

Let me know what do you think of the handling of the fitting procedure, as you can see in the [notebook I mentioned above](https://github.com/cosimoNigro/agnpy/blob/master/experiments/basic/sherpa_fitting_ssc_trial.ipynb) it takes few lines to define a `sherpa` model. 
We can open a new issue to discuss this.

PR #71, opened 2021-01-08
EC fit

jsitarek:
added a MWL SED of PKS1510-089 in 2015 (period B, EBL-deabsorbed)
added a tutorial to fit a combination of EC on DT + SSC to a FSRQ


cosimoNigro:
Thanks a lot Julian, it looks real good!

PR #73, opened 2021-01-13
Document spectral constraints

cosimoNigro:
I have written a little documentation for the `spectral_constraints` class.
I just put a little example, the API is very well documented with formulas. 
I also polished the notebook with the fit of PKS1510 SED and added it to the main doc pages.


cosimoNigro:
Partially address issue #73, now only tests are missing to close it.
I'll take care of them.

PR #81, opened 2021-02-19
Absorption mu s dependent

jsitarek:
absorption in DT for the case of gamma rays not moving along the jet
included also a few tests of the newly added methods


cosimoNigro:
Thanks a lot, @jsitarek, will check tomorrow!
Can I ignore PR #79? You modified the same lines in this PR. I am afraid if I merge #79 it will create a conflict (i.e. two PRs modifying the same line). Fine if I close it. Apologies for not checking #79 before.

jsitarek:
Hi @cosimoNigro 

sorry, probably I forked a new branch from already modified master (I should have made a dedicated branch for this fix in PR #79) and hence ended up with the same changes also in this branch

agnpy/absorption/absorption.py
```diff
@@ -350,6 +357,71 @@ def evaluate_tau_dt(
...
+        Returns
+        -------
+        :class:`~astropy.units.Quantity`
+            array of the SED values corresponding to each frequency
+        """
+        # conversions
+        epsilon_1 = nu_to_epsilon_prime(nu, z)
+        # multidimensional integration
+        # here uu is the distance that the photon traversed
+        uu = np.logspace(-5, 5, u_size) * r
```
cosimoNigro, line 409: Why do you start integrating from `1e-5 * r`, just for me to know because I always go from `r` to `1e5 * r` and maybe this is computationally better.

agnpy/absorption/absorption.py
```diff
@@ -386,6 +473,27 @@ def tau(self, nu):
         if isinstance(self.target, RingDustTorus):
             return self.tau_dt(nu)
 
+    def tau_mu_s(self, nu):
```
cosimoNigro, line 476: maybe we can include this in `tau` and call this function only when `mu_s != 0` or not `None`

agnpy/tests/test_absorption_mu_s.py
```diff
@@ -0,0 +1,116 @@
+# tests on the utils module
```
cosimoNigro, line 1: can we move this in the `test_absorption.py`? Just copying the entire `TestAbsorptionMuS` class over there

cosimoNigro: To have a single test unit per each sub-module.

cosimoNigro:
Dear @jsitarek,

thanks a lot for already taking care of this. 
I have left a small review, very minor comments only.

What I would like to ask you is to provide some written formulas with a sketch for the calculation you are introducing.
I could only verify `x_re_ring_mu_s` in the case in which the target and emitted photon are in the same plane, see sketch
photo5308020145937035691
Can you please provide something like this for the formulas you introduced in `utils.geometry`?
I do not want you to write them in latex, you can just send me a decent photo / scan of your calculation and I will take care of introducing them in the paper draft / docs.
Thank you!

agnpy/absorption/absorption.py
jsitarek: note that this 'u' (I called it in the code uu because u is often used for units) is not the same as 'l' that you are using, 'l' was a distance from the black hole, while 'u' is the distance from the photon injection place. If mu_s=1 then l=u+r, but in a general mu_s case there is a more complicated formula, and you would need to include du/dl if you integrate over l 
The integration in u should start from 0 (and go to infinity) and similar like in your mu_s=1 case it is better to do it in log scale, however then you have to be careful with the starting value. And note that if the photon starts far away from the reprocessing material even a small relative step in u can considerably change the distance and the angles, hence I start from 1.e-5 to be on a safe side w.r.t. all reasonable cases of absorption.

agnpy/absorption/absorption.py
jsitarek: makes sense, I can do the change

agnpy/tests/test_absorption_mu_s.py
jsitarek: sure, I kept it separated so we can decide how to organize the code, but I can move it there

jsitarek:
sure, I will provide the formulae (actually probably better that I write it in latex, my handwriting is terrible :-) )

agnpy/tests/test_absorption_mu_s.py
jsitarek: done (commit 51bd73c)

agnpy/absorption/absorption.py
jsitarek: I modified the function tau to have:
```
if self.mu_s==1: # default value
    ...
    if isinstance(self.target, RingDustTorus):
        return self.tau_dt(nu)
else:
    ...
    if isinstance(self.target, RingDustTorus):
        return self.tau_dt_mu_s(nu)
```
but then the test on test_abs_dt_vs_point_source fails and very likely because the point source behind the blob has the same problem with varying angle between photons along the integral 
I can make a similar function for computing tau for point source and arbitrary tau (the current one would have be be removed then, because it is only valid for mu_s =1 in which case for a point source we get tau=0, this is not a big problem.
However when I do so I'm pretty sure I will get the BLR vs point source failing as well, and computing the BLR absorption  is more tricky because I need to check some geometry relations first, so in the end I will have to disable at least that test

jsitarek: done in the latest commit (15d39d9).
as explained above I had to disable one of the tests, we can reenable it once the BLR absorption function is rewritten in a similar way (but most probably I can take care or it only next week, so if you are happy with the current version just merge it and I will make another PR for the BLR ).

cosimoNigro:
Thanks, looks good for now, I am merging.
If we want to change something we can open another PR. Thanks!

PR #85, opened 2021-03-10
Absorption BLR for arbitrary  mu_s

jsitarek:
functions for computing BLR absorption for the case of an arbitrary mu_s
attached below is the comparison with the old code for the case of mu_s ~1
tau_blr_on_axis_vs_general_r_10 _R_line_comparison
tau_blr_on_axis_vs_general_r_0 11_R_line_comparison
and updated notes with the formula also for BLR
[absorption_mu_s.pdf](https://github.com/cosimoNigro/agnpy/files/6117479/absorption_mu_s.pdf)
[absorption_mu_s.txt](https://github.com/cosimoNigro/agnpy/files/6117480/absorption_mu_s.txt)

the comparison with the point source behind the jet works fine and I reenabled it. While testing I saw some differences dependent on the binning used, but I will open a separate issue on this because I think this is a general problem.

### DIFF
--- a/agnpy/absorption/absorption.py
+++ b/agnpy/absorption/absorption.py
@@ -12,7 +12,9 @@ from ..utils.geometry import (
     mu_star_shell,
     x_re_ring,
     x_re_ring_mu_s,
+    phi_mu_re_shell,
     phi_mu_re_ring,
+    x_re_shell_mu_s,
 )
 from ..utils.conversion import nu_to_epsilon_prime, to_R_g_units
 from ..targets import PointSourceBehindJet, SSDisk, SphericalShellBLR, RingDustTorus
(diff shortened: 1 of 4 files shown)


cosimoNigro:
Hello Julian,
sorry for the delay, will find some time today to review it.

Cosimo

cosimoNigro:
Dear Julian,

sorry for the delay in reviewing it.
Everything looks ok in the code and I am merging it.

I have a pair of comment on the PDF with the formulas:
1) In the "spherical shell BLR" section, you forgot to specify how you compute the angles of the soft photon (theta and phi). You did that in the previous and following sections;
2) Eq. 9 should read `\sqrt(r^2 + u^2 \mu_s^2 + 2ru\mu_s)`, you forgot a `\mu_s` squared;
3) do you plan to implement the absorption for point-like source behind blob with an arbitrary `\mu_s` angle?

Thanks a lot, I will add these equations to the paper.

cosimoNigro:
I will correct these errors myself when copying your latex into the paper.

jsitarek:
> Dear Julian,
> 
> sorry for the delay in reviewing it.
> Everything looks ok in the code and I am merging it.
> 
> I have a pair of comment on the PDF with the formulas:
> 
>     1. In the "spherical shell BLR" section, you forgot to specify how you compute the angles of the soft photon (theta and phi). You did that in the previous and following sections;
> 

there is a sentence about this: 
"The direction angles are computed with the same Eq. 3 and 4."
the formula are the same, just the vector is different, you need to use the v_epsilon components from Eq 6 instead of Eq 1.


>     2. Eq. 9 should read `\sqrt(r^2 + u^2 \mu_s^2 + 2ru\mu_s)`, you forgot a `\mu_s` squared;
> 

I think the formula in the .pdf is fine, please note that 'u' has a different definition then your 'l', i.e. 'u' is the distance that photon travelled along the direction of movement (this is needed because you integrate over such 'u').
And then you get from cosine law:
x^2 = r^2 + u^2 - 2 * r * u * cos(pi-theta_s) 

>     3. do you plan to implement the absorption for point-like source behind blob with an arbitrary `\mu_s` angle?
> 

it is already there (I included it together with the previous PR on DT mu_s-dependent absorption)
check:
evaluate_tau_ps_behind_blob_mu_s

> 
> Thanks a lot, I will add these equations to the paper.

Issue #90, opened 2021-04-23
Calculation of BLR absorption fails for r = R_line


cosimoNigro:
The calculation of the absorption (and I presume also of the External Compton) on the `SphericalShellBLR` fails when `r=R_line`.

This is due to the definition of `x` (the distance from the reprocessing material to the blob), from Finke 2016:
`x^2 = r^2 + R_line^2 - 2 * r * R_re * mu_re `
where
- `r` =  height of the blob above the BH;
- `R_re` = radial coordinate of the reprocessing material;
- `mu_re` = zenith angle of the reprocessing material.

Clearly when `r=R_re`, `x=0` and the absorption (or EC) calculations gives `nan`, as the optical depth (or the SED) is dependent on `x^(-2)`.

Here a small snippet to trigger the issue:
```python
import numpy as np
import astropy.units as u
import astropy.constants as const
from agnpy.targets import SphericalShellBLR
from agnpy.absorption import Absorption

# BLR definition
L_disk = 2 * 1e46 * u.Unit("erg s-1")
xi_line = 0.024
R_line = 1.1 * 1e17 * u.cm
blr = SphericalShellBLR(L_disk, xi_line, "Lyalpha", R_line)

# absorption
abs_blr = Absorption(blr, r=R_line)

# tau
nu = np.logspace(22, 28) * u.Hz
tau_blr = abs_blr.tau(nu)

print(tau_blr)
```
returning:
```shell
[nan nan nan nan nan nan nan nan nan nan nan nan nan nan nan nan nan nan
 nan nan nan nan nan nan nan nan nan nan nan nan nan nan nan nan nan nan
 nan nan nan nan nan nan nan nan nan nan nan nan nan nan]
```
@jsitarek, shall we put a check in the [`x` calculation](https://github.com/cosimoNigro/agnpy/blob/master/agnpy/utils/geometry.py#L15)?

jsitarek:
Hi, I think this is the same thing which I mentioned in the second paragraph of #86 and this can also get triggered in other cases. 
I can make a fix, but there are a few different ways to do it:
1) set the value of integrant to 0 (or some bordering value) for the points where you end up with position ~R_line
2) shift the actual integration points that are ~R_line a bit e.g. 0.1 %)

I think the second solution is probably easier. Any preference?

cosimoNigro:
I agree, the second solution is easier, and probably also computationally faster.

jsitarek:
added a PR #91 that should solve this.
[...]

PR #91, opened 2021-04-28
added a special treatment for absorption on BLR if one of the grid points ends up very close to the shell

jsitarek:
also a test to check that it is working

this should solve problem reported  in issue #90


cosimoNigro:
Everything looks good.

PR #102, opened 2021-08-02
EC on CMB comparison, make uniform the SED format

cosimoNigro:
In this PR I have:
1. added a script to generate a SED for EC on CMB with jetset, to be compared against the same calculation by agnpy;
2. added this comparison to the automatic tests;
3. worked to standardise our MWL SED data format, they comply now with the [GADF specification](https://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html) and can be directly read via [Gammapy's `FluxPoints`](https://docs.gammapy.org/0.18.2/api/gammapy.datasets.FluxPointsDataset.html?highlight=fluxpoints#gammapy.datasets.FluxPointsDataset).


cosimoNigro:
I have also updated the tutorial notebooks to work with the new data and with some of the adjustments we made for the paper.
I will add also the gammapy's notebook once we finalise the paper content.

cosimoNigro:
Merging without review since the core library is not affected, only the test and docs.

PR #108, opened 2021-11-28
Some fixed after pylint check

pawel21:
I try open new PR for linting code.


pawel21:
I ran `pylint` for all files in `agnpy/agnpy/` and I tried impove code according to output from `pylint`.
Now most of the warnings are: `R0201: Method could be a function (no-self-use)` or e.g. ` Constant name "ps_test" doesn't conform to UPPER_CASE naming style (invalid-name)`, but to improve it would require quite deep code changes, I'm not sure if we want do it.

pawel21:
Hi @cosimoNigro What do you think about my code refactoring? What else I can do?

jsitarek (approved):
I had a quick look, and everything looks fine. There as a few functions that have the same docstring (test on disk properties), but I guess it is fine, the names of the functions are self-explanatory. @cosimoNigro shall we merge it?

cosimoNigro:
Thanks a lot Pawel, yes I also saw a lot of those complaints but I guess it is due to the naming scheme we used that is not totally conform to PEP 8 specifications. I saw that Julian already checked so I am merging. Thanks a lot!

PR #114, opened 2022-02-24
Update tutorial notebooks to use gammapy dev version

jepcor97:
This PR modifies the notebooks so they are able to run with gammapy dev version.

Changes in this PR:
 - Modifies test data `Mrk421_2011.ecsv` and `PKS1510-089_2015b.ecsv` removing points with repeeated energy since they could not be loaded by gammapy; as a workaround until this possibility is implemented.
 - Modifies the editting of uncertainty, since `FluxPoints` instances don't have an accesible `.data`attribute anymore, and it gave errors.
 - Small updates to the block of code doing the fitting.
 - Modifies the plot parameters (`kwargs` passed to plot) since the names had changed and they were not anymore compatible.


cosimoNigro:
Dear @jepcor97,

apologies for replying so late and thanks a lot for the effort of updating the notebooks.

I do not like the idea of deleting data points with repeated energies in the dataset.
Now, we had this discussion with @adonath while preparing the `agnpy_paper`.
If an additional info is available in the table (e.g. a string with the instrument name), Axel showed how to separate the flux points belonging to each instrument and then define a single `FluxPointsDataset` for each of them.
https://github.com/cosimoNigro/agnpy_paper/issues/18#issuecomment-982698805

This would solve the issue for the PKS1510-089 dataset, for which, thanks to Julian, we solved the problem of provenance.
There was a problem with single-bin flux measurement, but this will be fixed as soon as https://github.com/gammapy/gammapy/pull/3807 is merged in `gammapy-master`.

I contacted the authors of the Mrk421 2011 to ask if they had the MWL SED ordered by instrument, but did not receive any answer so far. I suggest we make an other attempt - I am writing to them now, and if we do not manage, then let us simply change the dataset! I am sure we can find another MWL SED, even from the same Mrk421.

As I said I do not like the idea of canceling measurements just because we cannot make flux points with them.
Also Andrea [is using the same dataset in `jetset`](https://jetset.readthedocs.io/en/1.1.2/user_guide/model_fit_1/Jet_example_model_fit.html), though he re-bins the flux points. I think it would be nice to keep the consistency between the two software examples.

Thanks for your work Juan.

cosimoNigro:
Since that dataset is "iconic" we can also consider keeping it and simply show an example of fit with `sherpa` using it.
We can differentiate the sources fitted with `Gammapy` and `sherpa`.

cosimoNigro:
@jepcor97, I added `Gammapy` and `sherpa` wrappers in a parallel PR #119.
This PR is superseded by it.

Thanks anyhow for your work.

PR #119, opened 2022-04-27
Incorporating the Gammapy wrapper in agnpy

cosimoNigro:
I think that some people are using our tutorial notebook for fitting with the Gammapy wrapper. 
https://agnpy.readthedocs.io/en/latest/tutorials/ssc_gammapy_fit.html

I don't think it is particularly pretty, especially because of the large cell of code defining the wrapper.
There was some discussion in one of the developer calls of `Gammapy` if we were to include the wrapper in `agnpy`, or in `Gammapy`, as currently done for the [`NaimaSpectralModel`](https://docs.gammapy.org/0.19/api/gammapy.modeling.models.NaimaSpectralModel.html).

I think it is ok to have the wrapper within `agnpy`, we will be introducing `Gammapy` as an additional dependency but I think it's also fine. Now we will have a centralised wrapper, and hopefully adding a line of code we can solve the issue encountered by @guillaumegrolleron and @jepcor97 when fitting and simulating with the latest `Gammapy` versions (see the comments in the wrappers).

I thought about having a **single wrapper per each radiative processes** but this implies:
1. re-writing the same code several times (one per each radiative processes);
2. users having to link all the different parameters between the two (or more) radiative processes they are considering and then define yet another `Gammapy.SpectralModel` summing them.

So I implemented **a wrapper per each scenario / source type**, I have written:
1. one with synchrotron + SSC for BL Lacs;
2. one with synchrotron + SSC + external Compton (on BLR and DT) for FSRQs (the EC on BLR computation can be switched off as it is time consuming).

Parameters are read from an instance of the emission region and target.

Many things are missing, still I wanted to open the review to start to discuss the design and the organisation of the wrapper.
@jsitarek what do you think? Let me know what you think about the strategy, the naming scheme, the position of the new submodule.

I will also be happy to have the opinions of @QRemy and @atreyeeS, from the `Gammapy` side - if they have time.

Follows a script to cross-check the normal computation against the new wrappers implemented (we can incorporate this in the tests).

```python
import numpy as np
import astropy.units as u
from astropy.coordinates import Distance
from agnpy.emission_regions import Blob
from agnpy.targets import SphericalShellBLR, RingDustTorus
from agnpy.synchrotron import Synchrotron
from agnpy.compton import SynchrotronSelfCompton, ExternalCompton
from agnpy.wrappers import SpectralModelSSC, SpectralModelEC
from agnpy.utils.plot import sed_x_label, sed_y_label
import matplotlib.pyplot as plt

# emission region and targets
# - SSC blob
spectrum_norm = 1e48 * u.Unit("erg")
spectrum_dict = {
    "type": "PowerLaw",
    "parameters": {
        "p": 2.8,
        "gamma_min": 1e2,
        "gamma_max": 1e7
    }
}
R_b = 1e16 * u.cm
B = 1 * u.G
z = Distance(1e27, unit=u.cm).z
delta_D = 10
Gamma = 10
ssc_blob = Blob(R_b, z, delta_D, Gamma, B, spectrum_norm, spectrum_dict)

# - EC blob
spectrum_norm = 6e42 * u.erg
parameters = {
    "p1": 2.0,
    "p2": 3.5,
    "gamma_b": 1e4,
    "gamma_min": 20,
    "gamma_max": 5e7,
}
spectrum_dict = {"type": "BrokenPowerLaw", "parameters": parameters}
R_b = 1e16 * u.cm
B = 0.56 * u.G
z = 1
delta_D = 40
Gamma = 40
ec_blob = Blob(R_b, z, delta_D, Gamma, B, spectrum_norm, spectrum_dict)

# - BLR definition
xi_line = 0.024
L_disk = 1e46 * u.Unit("erg s-1")
R_line = 1e17 * u.cm
blr = SphericalShellBLR(L_disk, xi_line, "Lyalpha", R_line)
# - dust torus definition
T_dt = 1e3 * u.K
csi_dt = 0.1
dt = RingDustTorus(L_disk, csi_dt, T_dt)

# distance
r = 1e16 * u.cm

# agnpy radiative processes
synch = Synchrotron(ssc_blob)
ssc = SynchrotronSelfCompton(ssc_blob)

synch_ec = Synchrotron(ec_blob)
ssc_ec = SynchrotronSelfCompton(ec_blob)
ec_blr = ExternalCompton(ec_blob, blr, r)
ec_dt = ExternalCompton(ec_blob, dt, r)

# radiative processes wrapped by Gammapy
ssc_model = SpectralModelSSC(ssc_blob)
ec_model = SpectralModelEC(ec_blob, blr, dt, r, ec_blr=False)

# SED computation
nu = np.logspace(9, 29, 100) * u.Hz
E = nu.to("eV", equivalencies=u.spectral())

# SSC model
sed_ssc_agnpy = synch.sed_flux(nu) + ssc.sed_flux(nu)
sed_ssc_gammapy = (E**2 * ssc_model(E)).to("erg cm-2 s-1")

# EC model
sed_ec_agnpy = (
    synch_ec.sed_flux(nu) + ssc_ec.sed_flux(nu) + ec_dt.sed_flux(nu)
)
sed_ec_gammapy = (E**2 * ec_model(E)).to("erg cm-2 s-1")

# SSC comparison
plt.loglog(nu, sed_ssc_agnpy, ls="-", label="SSC, agnpy")
plt.loglog(nu, sed_ssc_gammapy, ls="--", label="SSC, Gammapy wrapper")
plt.xlabel(sed_x_label)
plt.ylabel(sed_y_label)
plt.ylim([1e-13, 1e-9])
plt.legend()
plt.show()

# EC comparison
plt.loglog(nu, sed_ec_agnpy, ls="-", label="EC, agnpy")
plt.loglog(nu, sed_ec_gammapy, ls="--", label="EC, Gammapy wrapper")
plt.xlabel(sed_x_label)
plt.ylabel(sed_y_label)
plt.ylim([1e-18, 1e-14])
plt.legend()
plt.show()
```
Figure_1
Figure_2


agnpy/wrappers/gammapy.py
```diff
@@ -0,0 +1,208 @@
+from multiprocessing.sharedctypes import Value
+from xml.etree.ElementInclude import XINCLUDE_INCLUDE
+import astropy.units as u
+from ..synchrotron import Synchrotron
+from ..compton import ExternalCompton
+from ..compton import SynchrotronSelfCompton
+from gammapy.modeling.models import SpectralModel, Parameter, Parameters
+
+
+class SpectralModelSSC(SpectralModel):
```
QRemy, line 10: Maybe the wrapped models can follow the naming conventions of gammapy models with SpectralModel at the end :
`SynchrotronSelfComptonSpectralModel`, `ExternalComptonSpectralModel`

agnpy/wrappers/gammapy.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # emission region parameters
+        z = Parameter("z", blob.z)
+        d_L = Parameter("d_L", blob.d_L)
+        delta_D = Parameter("delta_D", blob.delta_D)
+        B = Parameter("B", blob.B)
+        R_b = Parameter("R_b", blob.R_b)
+        parameters.extend([z, d_L, delta_D, B, R_b])
+
+        self.default_parameters = Parameters(parameters)
```
QRemy, line 38: gammapy models parameters are usually defined at class level and come with default values, but I guess you prefer to avoid this because it's hard to define good defaults here.

agnpy/wrappers/gammapy.py
```diff
@@ -0,0 +1,208 @@
...
+        # emission region parameters
+        z = Parameter("z", blob.z)
+        d_L = Parameter("d_L", blob.d_L)
+        delta_D = Parameter("delta_D", blob.delta_D)
+        B = Parameter("B", blob.B)
+        R_b = Parameter("R_b", blob.R_b)
+        mu_s = Parameter("mu_s", blob.mu_s)
+        # distance of the emission region along the jet
+        r = Parameter("r", r)
+        parameters.extend([z, d_L, delta_D, B, R_b, mu_s, r])
```
QRemy, line 114: maybe adding something like
self.emission_region_parameters
self.targets_parameters
could help to simplify the sorting of the parameters later in evaluate
(and could also be convenient for user that want to look only at one group of parameters)

AtreyeeS:
This should be quite useful @cosimoNigro . One general comment here... Currently, gammapy `FluxPointEstimator` needs either an `amplitude` or `norm` parameter to be defined on the model. (I imagine that it would be interesting to fit reduced datasets directly with physical models and extract flux points). This is not the case here.

It is not a blocker, as the user can still multiply `SpectralModelSSC` by a normed model to the specific case of extracting flux points, but it is probably important to be correctly conveyed

AtreyeeS:
Sorry my previous comment was not totally correct. As per the latest developments, users need to specify `is_norm` for scalable parameters (see discussions in https://github.com/gammapy/gammapy/issues/3690#issuecomment-1116218619)

So one option (to simplify things for the users) here maybe be to add a `norm` parameter which must be kept frozen, but can be used internally during flux point estimation...

agnpy/wrappers/gammapy.py
cosimoNigro: Done, renamed following the `Gammapy` naming scheme.

agnpy/wrappers/gammapy.py
cosimoNigro: Yes, it is a bit difficult to establish defaults, in general. Additionally each emission region comes with its own electron distribution of which I do not know the parameters. I added functions to fetch these parameters from the physical objects and set some reasonable ranges of values.

agnpy/wrappers/gammapy.py
cosimoNigro: I grouped the parameters following your suggestions, I don't know though how to re-use this grouping in the `evaluate` function. I see that the `NaimaSpectralModel` in `Gammapy` does something similar, i.e. uses in the `evaluate` some attribute of the class, but then this `_update_naima_parameters` is needed https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/models/spectral.py#L2107. Let me know if you have some suggestion here.

cosimoNigro:
Thanks for your comments @AtreyeeS, now `k_e`, the normalisation of the electron distribution has `is_norm=True` when converted to a `gammapy.modeling.Parameter`. The flux or SED should scale linearly with the normalisation of the radiating electrons.

cosimoNigro:
Will add some tests later.

AtreyeeS:
>  The flux or SED should scale linearly with the normalisation of the radiating electrons, as it is just a multiplicative factor in all the SED calculations.

I think that while the synchrotron and EC flux goes as `k_e`, the SSC goes as `k_e^2` (because its the same population of electrons, so one factor of k_e from the electrons, and one from the sync photons). 
You can see the same in the fig below:

Unknown

cosimoNigro:
Yes, you are totally right! Thank you!
I will then just introduce a general `amplitude` multiplying the SED and that would be the `norm` of the model.
Would that be ok.

AtreyeeS:
@cosimoNigro : In gammapy, we tend to use `amplitude` as a dimensional parameter, and `norm` as a scaling, so maybe `norm` is better? And document it well that the `norm` must be **frozen** during the fit. I don't know if it is possible to hide it from the users in any way, but if you are separating parameters as `blob_parameters` and `jet_parameters` as @QRemy suggested, then at least the norm should not show up in either of those

cosimoNigro:
Ok, so I added the `norm`, and I put it a the bottom of the parameter list. It does appear if you call all the parameters together (don't know how to hide it).

```python
from agnpy.emission_regions import Blob
from agnpy.wrappers import SynchrotronSelfComptonSpectralModel

blob = Blob()
ssc_model = SynchrotronSelfComptonSpectralModel(blob)

print(ssc_model.parameters.to_table())
```
```shell
  type      name     value    unit   error      min       max    frozen is_norm link
-------- --------- ---------- ---- --------- --------- --------- ------ ------- ----
spectral       k_e 9.2748e+06 cm-3 0.000e+00 1.000e+00 1.000e+09  False   False     
spectral         p 2.8000e+00      0.000e+00 1.000e+00 5.000e+00  False   False     
spectral gamma_min 1.0000e+02      0.000e+00 1.000e+00 1.000e+03   True   False     
spectral gamma_max 1.0000e+07      0.000e+00 1.000e+04 1.000e+08   True   False     
spectral         z 6.9000e-02      0.000e+00 1.000e-03 1.000e+01   True   False     
spectral       d_L 9.9203e+26   cm 0.000e+00 1.368e+25 3.271e+29   True   False     
spectral   delta_D 1.0000e+01      0.000e+00 1.000e+00 1.000e+02  False   False     
spectral         B 1.0000e+00    G 0.000e+00 1.000e-04 1.000e+03  False   False     
spectral       R_b 1.0000e+16   cm 0.000e+00 1.000e+12 1.000e+18  False   False     
spectral      norm 1.0000e+00      0.000e+00 1.000e-01 1.000e+01   True    True   
```

As @AtreyeeS said, it does not appear in the "grouped" parameters
```python
print(ssc_model.spectral_parameters.to_table())
```
```shell
  type      name     value    unit   error      min       max    frozen is_norm link
-------- --------- ---------- ---- --------- --------- --------- ------ ------- ----
spectral       k_e 9.2748e+06 cm-3 0.000e+00 1.000e+00 1.000e+09  False   False     
spectral         p 2.8000e+00      0.000e+00 1.000e+00 5.000e+00  False   False     
spectral gamma_min 1.0000e+02      0.000e+00 1.000e+00 1.000e+03   True   False     
spectral gamma_max 1.0000e+07      0.000e+00 1.000e+04 1.000e+08   True   False
```

```python
print(ssc_model.spectral_parameters.to_table())
```
```shell
  type     name    value    unit   error      min       max    frozen is_norm link
-------- ------- ---------- ---- --------- --------- --------- ------ ------- ----
spectral       z 6.9000e-02      0.000e+00 1.000e-03 1.000e+01   True   False     
spectral     d_L 9.9203e+26   cm 0.000e+00 1.368e+25 3.271e+29   True   False     
spectral delta_D 1.0000e+01      0.000e+00 1.000e+00 1.000e+02  False   False     
spectral       B 1.0000e+00    G 0.000e+00 1.000e-04 1.000e+03  False   False     
spectral     R_b 1.0000e+16   cm 0.000e+00 1.000e+12 1.000e+18  False   False  
```

I integrated the computation of the SED with the `Gammapy` wrappers in the tests, I find quite some discrepancy at the highest energies. I expect exactly identical results as I am using the same functions. Will try to dig deeper later.

gammapy_ssc_wrapper

gammapy_ec_wrapper

cosimoNigro:
Ok, very well,the test were failing because `Gammapy` - that I am now installing - uses `astropy 5.0` that has dropped support for python 3.7. Just removed python 3.7 from the tests.

cosimoNigro:
Will take time to make a final couple of improvements in the EC wrapper and then merge.

adonath:
> Is it because I am now using several FluxPointsDatasets?

Theoretically yes. At some point the Python loop in `Datasets.stat_sum()` could become the bottleneck, but I doubt this the case here. The integrations performed within the loop should still be the dominant part of the computation.

Can you compare the number of fit iterations for both cases? Is it the same? My strong guess here is no. It seems the convergence changed here...

adonath:
The total fit on my M1 machine takes 12 min with 2354 iterations. The number of iterations seems rather high to me...

QRemy:
The  1-d joint fit of FluxPointsDatasets is not covered by gammapy-benchmark so I don't have the history of evaluation time to compare.
Could you try to run a profiler to see what takes most of the time, for example :
```
pip install pyinstrument
pyinstrument -r html -o check_gammapy_fit.html check_gammapy_fit.py
```

cosimoNigro:
> Could you try to run a profiler to see what takes most of the time, for example

Done, the most time-consuming tasks seem to be those related with optimisation and flux points. The full .html is attached
Screenshot from 2022-06-20 16-21-51

full .html [here](https://gist.github.com/cosimoNigro/697d5e31d46eb5c51328ddad8be86282).

AtreyeeS:
> Is it because I am now using several FluxPointsDatasets?

Maybe it can contribute a part of the reason. I tried a very simple test - fit the hawk crab flux points supplied within gammapy data 

- as one `FluxPointsDataset`
- broken into 10 `FluxPointsDataset`

The fit results do not change, but the time for fitting increased by a factor of 7, you can see more here: https://github.com/AtreyeeS/test_notebooks/blob/master/test_joint_fit.ipynb

QRemy:
```
from agnpy.utils.math import log
%timeit log(np.ones(10))
%timeit [log(np.ones(1)) for k in range(10)]
```
Arrays are just more efficient...
Here, the increase in time is proportional to the number of call to the log function,
so indeed more datasets with less flux points per dataset can explain the  difference. 
The FluxPointsDatasets should probably concatenate the data internally to run as fast as a single FluxPointDataset.

adonath:
@cosimoNigro, @QRemy and @AtreyeeS  I agree it seems the increase in computation time here seems to come from splitting the datasets. However all things considered I think it's still the right way to go:

- This case is special in the sense that it only includes flux point datasets and a rather high number of them. I guess once one adds a single `MapDataset` for Fermi-LAT or a IACT it will dominate the computation time anyway... 
- Looking at the example, the problem is not the computation time, but rather the model is dominated by systematics. The fit runs >2000 iterations but still does not fit all data points, especially at the low energy end. So the model is well constrained but does not fit the data well.
- So it's rather important to add weights and possibly energy scale systematics. I think the latter can only be achieved in a clean way by splitting into datasets and assigning an energy scale parameter to each of them.

> The FluxPointsDatasets should probably concatenate the data internally to run as fast as a single FluxPointDataset.

I'm not sure it is worth it. From the Gammapy side we could rather work on:
- Support for non-contiguous energy axes, such that flux points can be merged again
- Support for parallelization of datasets to improve performance.

But those are bigger development projects...

cosimoNigro:
Yes @adonath, thanks for asking! I will send to the Gammapy devs the link to the docs with the updated notebooks, once I finish everything. The speed of the fit is now 3 minutes, compared to the 2 minutes I had with `Gammapy 0.18.2`, so everything looks good.

cosimoNigro:
Ok, I am done with this.
I have introduced wrappers both for `Gammapy` and `sherpa` models.
I have also added some helper functions to load the SED data in a `sherpa.data.Data1D` object or in a list of `gammapy.datasets.FluxPointsDatasets`.
Proper tests have been added for the new classes / functions added.
I can now obtain the same results from the two wrappers for the two sources considered, Mrk421 and PKS1510-089.

I will make a new release after I merge.

PR #124, opened 2022-08-17
Introduce protons energy distributions

cosimoNigro:
This pull requests introduces the possibility to have protons energy distributions in `agnpy`.
This is the first step in the introduction of hadronic models in the package.

I thought that the less API-breaking way of introducing protons energy distributions was to allow for a `mass` parameter in our spectra. If the electron mass is specified, then we get an electrons energy distribution, if a proton mass is specified, then we get a protons energy distribution. Example:

```python
import astropy.units as u
from astropy.constants import m_e, m_p
from agnpy.spectra import PowerLaw

# if the mass parameter the default mass=m_e will be used
n_e = PowerLaw(k=1e-2 * u.Unit("cm-3"), p=2.1, gamma_min=1e2, gamma_max=1e6)
n_p = PowerLaw(k=1e-2 * u.Unit("cm-3"), p=2.1, gamma_min=1e2, gamma_max=1e6, mass=m_p)
print(n_e)
print(n_p)
```
 which returns
```
* electrons energy distribution
 - power law
 - k: 1.00e-02 1 / cm3
 - p: 2.10
 - gamma_min: 1.00e+02
 - gamma_max: 1.00e+06

* protons energy distribution
 - power law
 - k: 1.00e-02 1 / cm3
 - p: 2.10
 - gamma_min: 1.00e+02
 - gamma_max: 1.00e+06
```
The particles distributions can be initialised also from the total particle density, the total energy density or the total energy.
These functionalities were before handled by the `Blob`, and I think it is better to have them now in the particles distributions.

How are then the particles distributions passed to the `Blob`? Simple, with `n_e` and `n_p` parameters. Continuing from the previous snippet:
```python
from agnpy.emission_regions import Blob

blob = Blob(n_e=n_e)
print(blob)
```
```
* Spherical emission region
 - R_b (radius of the blob): 1.00e+16 cm
 - t_var (variability time scale): 4.13e-01 d
 - V_b (volume of the blob): 4.19e+48 cm3
 - z (source redshift): 0.07
 - d_L (source luminosity distance):9.92e+26 cm
 - delta_D (blob Doppler factor): 1.00e+01
 - Gamma (blob Lorentz factor): 1.00e+01
 - Beta (blob relativistic velocity): 9.95e-01
 - theta_s (jet viewing angle): 5.74e+00 deg
 - B (magnetic field tangled to the jet): 1.00e+00 G
 - xi (coefficient for 1st order Fermi acceleration) : 1.00e+00
* electrons energy distribution
 - power law
 - k: 1.00e-02 1 / cm3
 - p: 2.10
 - gamma_min: 1.00e+02
 - gamma_max: 1.00e+06
```

As we can see, if no proton distribution is specified, then we only have leptons, we can always add the proton distribution
```python
blob.n_p = n_p
print(blob)
```
```
* Spherical emission region
 - R_b (radius of the blob): 1.00e+16 cm
 - t_var (variability time scale): 4.13e-01 d
 - V_b (volume of the blob): 4.19e+48 cm3
 - z (source redshift): 0.07
 - d_L (source luminosity distance):9.92e+26 cm
 - delta_D (blob Doppler factor): 1.00e+01
 - Gamma (blob Lorentz factor): 1.00e+01
 - Beta (blob relativistic velocity): 9.95e-01
 - theta_s (jet viewing angle): 5.74e+00 deg
 - B (magnetic field tangled to the jet): 1.00e+00 G
 - xi (coefficient for 1st order Fermi acceleration) : 1.00e+00
* electrons energy distribution
 - power law
 - k: 1.00e-02 1 / cm3
 - p: 2.10
 - gamma_min: 1.00e+02
 - gamma_max: 1.00e+06
* protons energy distribution
 - power law
 - k: 1.00e-02 1 / cm3
 - p: 2.10
 - gamma_min: 1.00e+02
 - gamma_max: 1.00e+06
```

This way the `Blob` keeps having a `Blob.n_e` attribute which is the callable function corresponding to the electron distribution. This attribute is already called by all leptonic radiative processes. Similarly, when we start to implement hadronic radiative processes, we can invoke `Blob.n_p`.

Note I have removed the previous blob initialisation using a dictionary for the particle distribution. In hindsight, it could have been better implemented from the beginning. The different initialisation method (from total density, total energy) are to be used directly when defining the particle distribution.

So far, I modified the `spectra`, `emission_regions` and `synchrotron` modules to work with the new definition of the particle class. Note only one line was changed in `synchrotron.py` to use this new particles definition.

I still have to modify the `compton` and `fit` modules, the other modules should be unaffected by this major change.

P. S. the only snippet in the `agnpy` paper keeps working with this new definition.

@jsitarek, I am assigning you for review, but we can discuss the changes in a call, since they are many.

@IlariaViale, @dimaniad6, after this is merged, you can start working on the proton synchrotron.


cosimoNigro:
I finished integrating the new particles distribution in all the classes.
Locally all the tests are passing.

I would open a follow-up PR to update the documentation as this one comes already with a lot of changes.

cosimoNigro:
After the tests run I will merge.

Issue #122, opened 2022-08-11
Add sherpa and gammapy among the dependencies (add them to install_requires in setup.py)


cosimoNigro:
So far we had only `numpy`, `scipy`, `astropy` and `matplotlib` among the dependencies.
These are specified in `setup.py` and automatically downloaded and installed when `agnpy` is installed via `pip` or `conda`.

After #119, I have introduced `gammapy` and `sherpa` wrappers for the `agnpy` radiative models.
This means that there are now classes in the new `agnpy.fit` submodule that are built on `sherpa` and `gammapy` objects.

Trying to simulate an installation from pip, I created a new conda environment.
```
mamba create -n agnpy_0.2.0_test python=3.9 
conda activate agnpy_0.2.0_test
```
I created a distribution (using the same command used by [the github action automatically publishing our releases on PyPI](https://github.com/cosimoNigro/agnpy/blob/master/.github/workflows/pip-upload.yml#L28)) and installed it
```
python setup.py sdist bdist_wheel
cd dist
pip install agnpy-0.2.0.tar.gz
```

When I try to import any object I get the following error:

```python
>>> from agnpy.emission_regions import Blob
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/cosimo/software/miniconda3/envs/agnpy_0.2.0_test/lib/python3.9/site-packages/agnpy/__init__.py", line 9, in <module>
    from .fit import *
  File "/Users/cosimo/software/miniconda3/envs/agnpy_0.2.0_test/lib/python3.9/site-packages/agnpy/fit/__init__.py", line 1, in <module>
    from .models import *
  File "/Users/cosimo/software/miniconda3/envs/agnpy_0.2.0_test/lib/python3.9/site-packages/agnpy/fit/models.py", line 1, in <module>
    from .gammapy_wrapper import (
  File "/Users/cosimo/software/miniconda3/envs/agnpy_0.2.0_test/lib/python3.9/site-packages/agnpy/fit/gammapy_wrapper.py", line 6, in <module>
    from gammapy.modeling import Parameter, Parameters
ModuleNotFoundError: No module named 'gammapy.modeling'
```

Curiously, just repeating the import a second time in the same python terminal does not raise any error
```python
>>> from agnpy.emission_regions import Blob
>>> blob = Blob()
>>> print(blob)
* spherical emission region
 - R_b (radius of the blob): 1.00e+16 cm
[...]

PR #125, opened 2022-08-18
Fix gammapy and sherpa imports

cosimoNigro:
This PR is a followup of issue #122.

I thought that adding `gammapy` and `sherpa` to the `install_requires` of `conf.py` would solve the issue.
Instead when I try to create a distribution file and install it, the dependencies cannot be resolved via `pip`, that fails to install `sherpa`. The problem is not there if I install the packages with `mamba` after installing the dist file.

Therefore, following the initial suggestion of @jsitarek, I added a `try/except` clause in `agnpy/fit/__init__.py` printing a warning message if the import fails. The rest of the modules can be used then, even without `gammapy` and `sherpa`.

I added a clarification in the docs suggesting to install the fitting packages separately if `agnpy` is installed via `pip` or `conda`.

@jsitarek can you take a quick look?

Solves #122.


cosimoNigro:
It seems to be working for the tests I could do, so I would merge and make the release.

PR #126, opened 2022-08-23
Fixes in the docs before 0.2.0 release

cosimoNigro:
This PR implements minor fixes in the documentation before making the release 0.2.0.

Among them I clarified, both in the README and in the main doc page, how to cite agnpy.

After thinking about it, I think this is the best option for acknowledging, the [CITATION.cff](https://citation-file-format.github.io/) files looks more useful in case a proper release paper is absent, and one wants to cite the zenodo record. I specify in the docs and in the README page to cite both.

I also removed from the docs this page where all the calculations are derived, it was updated, and the same formulas can now be found in the agnpy release paper.


cosimoNigro:
One last thing: the notebooks included in the docs can be automatically run by sphinx, so from now on we should submit them without the output cells, i.e. non-executed. This way they will be easier to review and we can reduce the amount of lines we add each time. 

The only exceptions are the fitting notebooks, as some of them take a while to run, I suggest we commit those already executed.

PR #128, opened 2022-09-29
Introducing the InterpolatedDistribution subclass at spectra.py

dimaniad6:
We add an interpolation subclass of the ElectronDistribution to the spectra.py. The class receives as an input 2 arrays, one with the gammas and the other with the particle densities, performs the interpolation in the logarithmic space and returns back the function. The tool used was CubicSpline from scipy. 

The corresponding pytests have been added to the test_spectra.py: data are generated from the 4 already existing distributions of agnpy (power law, log parabola, exp cut off, broken power law) and then are interpolated. The interpolated function for each case is being compared to the data, by computing the deviation of each data point to the point given by the function. 
The class also calculates the SSA integrand. For the pytests, the % deviations that are accepted are less than 0.001% for the interpolation function, and less than 1% for the SSA.  

*Some important comments:*

The comparisons between the SSA integrand expected values and the values from the interpolation have very small deviations, small enough to pass the pytests, but there's a 10% deviation in the case of the broken power law distribution just at the point of the discontinuity, where the break is at. That is because CubicSpline smooths out the discontinuities, resulting in relatively large deviations, enough to be seen when calculating the derivative. An important note here is that such deviations maybe have an noticeable effect not just on the derivative, but also on the interpolation function itself. They were not 'seen' by the pytests, but perhaps a better pytest should be constructed, in order to understand if there are big deviations on points that do not have the exact same gammas as the ones from the data (from which the interpolation function was derived), at the areas of such discontinuities.

### DIFF
--- a/agnpy/spectra/spectra.py
+++ b/agnpy/spectra/spectra.py
@@ -13,7 +13,7 @@ __all__ = [
     "ExpCutoffPowerLaw",
     "BrokenPowerLaw",
     "LogParabola",
-    "Interpolation"
+    "InterpolatedDistribution"
 ]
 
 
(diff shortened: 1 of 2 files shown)


agnpy/spectra/spectra.py
```diff
@@ -473,3 +477,37 @@ def __str__(self):
...
+            0
+        )
+
+    def __call__(self, gamma):
+        return self.evaluate(
+            gamma, self.gamma_min, self.gamma_max
+        )
+
+    #deriv is the derivative of 10**(f_log(np.log10(gamma))). Chain rule is used
+    def SSA_integrand(self, gamma):
```
cosimoNigro, line 510: can you double-check the formula? This is what I obtain and I see that at least a ln(10) is missing

CodeCogsEqn

Also why are you using each time `np.power(10,self.f_log(np.log10(gamma)))` to evaluate the actual function, when you have that defined in `evaluate` already?

agnpy/tests/test_spectra.py
```diff
@@ -86,6 +93,76 @@ def broken_power_law_times_gamma_integral(k_e, p1, p2, gamma_b, gamma_min, gamma
...
 
+# Data are generated, following the four already implemented distributions of agnpy.
+# The interpolation function is derived for each data set and then is interpolated and
+# data points are being compared:
+
+def pwl_data(k_e_test,p_test,gamma_min_test,gamma_max_test):
+
+    pwl_data = np.zeros((2,100),float)
+    gamma1 = np.logspace(np.log10(gamma_min_test),np.log10(gamma_max_test),100)
+    pwl_data[0,:] = gamma1
```
cosimoNigro, line 104: Why are you creating this matrix of 2 * 100?
Can't you pass directly the array (which is always the same, `gamma1`) and then the values generated by `pwl_test(gamma1)` - or `bpwl_test(gamma1)` - to your function?

agnpy/spectra/spectra.py
```diff
@@ -473,3 +477,37 @@ def __str__(self):
...
+
+class Interpolation(ElectronDistribution):
+
+    def __init__(
+        self,
+        gamma_data,
+        n_data
+    ):
+        self.gamma_data = gamma_data
+        self.n_data = n_data.value
```
cosimoNigro, line 490: Maybe add a check that the distribution is provided in `cm-3`? Or convert to that value?

agnpy/tests/test_spectra.py
```diff
@@ -86,6 +93,76 @@ def broken_power_law_times_gamma_integral(k_e, p1, p2, gamma_b, gamma_min, gamma
...
+    ep_data[1,:] = epwl_test(gamma1).value
+
+    return ep_data
+
+#power law data and interpolation
+pwl_data = pwl_data(
+    k_e_test,p_test,gamma_min_test,gamma_max_test
+)
+pwl_inter = Interpolation(
+    pwl_data[0,:], pwl_data[1,:]*u.Unit('cm-3')
```
cosimoNigro, line 143: see my comment before, this could be:
```python
Interpolation(gamma1, pwl_test(gamma1))
```

agnpy/tests/test_spectra.py
```diff
@@ -407,3 +484,82 @@ def test_from_norm_at_gamma_1(self):
...
+        assert not np.all(values[~condition])
+
+    """ 2. assert that the interpolated function does not have large deviations from the original one """
+
+    def test_power_vs_inter(self):
+        a = abs(pwl_inter(pwl_data[0,:]).value - pwl_data[1,:])/pwl_data[1,:]
+        assert max(a) < 1e-5
+
+    def test_bpwl_vs_inter(self):
+        a = abs(bpwl_inter(bpwl_data[0,:]).value - bpwl_data[1,:])/bpwl_data[1,:]
```
cosimoNigro, line 527: [numpy.allclose](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html) and [units.allclose](https://docs.astropy.org/en/stable/api/astropy.units.allclose.html) already perform this calculation for you. One does it with simple arrays, the other with array of units.

cosimoNigro: Same for the following checks.

cosimoNigro (changes requested):

cosimoNigro:
Hello @dimaniad6, thanks a lot for your work!
The PR looks already good, I just left some comments to simplify it.

Maybe let us change the name of the new class from `Interpolation` to `InterpolatedDistribution`, or something on this line?

Thanks!

PR #132, opened 2023-01-09
Introducing Proton Synchrotron 

dimaniad6:
We introduce the hadronic process of proton synchrotron radiation, after the changes of Cosimo that added the possibility of a proton particle distribution. The code is basically the same as the electron synchrotron, but instead of the electron mass, the proton mass is used. To do that, we introduced the new mass variable, which can take two values, m_e and m_p.  With this PR, we also added an additional distribution of a broken power law with an exponential cut-off. The corresponding pytests are included.


IYNBI:
Hi

I have just started to use agnpy and gammapy for my project.
Can you help me figure out why this is showing up this error?

Regards
Inbisat Yousuf

cosimoNigro:
@jsitarek would you have time to take a look at @dimaniad6 PR?

agnpy/synchrotron/proton_synchrotron.py
```diff
@@ -0,0 +1,183 @@
...
+        d_L,
+        delta_D,
+        B,
+        R_b,
+        n_p,
+        *args,
+        integrator=np.trapz,
+        gamma=gamma_e_to_integrate,
+    ):
+        """Computes the syncrotron self-absorption opacity for a general set
```
jsitarek, line 50: synchrotron

agnpy/synchrotron/proton_synchrotron.py
```diff
@@ -0,0 +1,183 @@
...
+e = e.gauss
+B_cr = 4.414e13 * u.G  # critical magnetic field
+
+class ProtonSynchrotron:
+    """Class for synchrotron radiation computation
+
+    Parameters
+    ----------
+    blob : :class:`~agnpy.emission_region.Blob`
+        emitting region and electron distribution
```
jsitarek, line 21: electrons or protons?

agnpy/utils/conversion.py
```diff
@@ -1,24 +1,37 @@
...
+            (u.Hz, u.Unit(""), lambda x: h.cgs * x / mpc2, lambda x: x * mpc2 / h.cgs)
+            ]
+    else:
+        raise ValueError(
+            f"Provide either the electron or the proton mass. "
+        )
+    return epsilon_equivalency
+
+
+def nu_to_epsilon_prime(nu, z=0, delta_D=1, m = m_e):
```
jsitarek, line 30: I think this is dangerous. Keeping two different conversions of nu to E depending on the particle can cause problems if somebody would like to plot at the same time electron and proton emission

jsitarek (commented):
sorry for a late feedback. 
I have two main comments:
1) synchrotron-self-absorption was introduced also for proton synchrotron radiation, basically by substituting m_e for m_p. The usual formula that is used for SSA is approximate: the slope is derived for the strong absorption boardering case and then crossing of the two boardering cases is used to get the full formula. Maybe it is just my ignorance, but I do not remember ever seeing SSA being considered for protons. The proton-photon interactions have a much smaller cross-section than electron-photon, which should make it much less important process.  
2) a new equivalency was introduced that changes the frequency to energy conversion depending on the particle mass. It might be fine, but I am afraid that it might also produce problems when user calculates both electron and proton processes and wants to put them in the same plot

agnpy/synchrotron/proton_synchrotron.py
dimaniad6: Thank you for the comments. The reason we put SSA was mostly because Cerruti also included it on his code for proton synchrotron (unless we interpreted wrong the plot from his code). Since it is not relevant, do you think we should exclude it for protons?

jsitarek: if you mimick Matteo's code it is fine to keep it as it is, but you should include a reference to the original code in the docstring of ProtonSynchrotron class

agnpy/spectra/spectra.py
```diff
@@ -565,6 +566,131 @@ def __str__(self):
         )
 
 
+class ExpCutoffBrokenPowerLaw(ParticleDistribution):
+    r"""Class describing an exponential cutoff broken power-law particle distribution.
+    When called, the particle density :math:`n(\gamma)` in :math:`\mathrm{cm}^{-3}` is returned.
+
+    .. math::
+        n(\gamma') = k \left[
+        \left(\frac{\gamma'}{\gamma'_b}\right)^{-p_1} exp(-\gamma'/\gamma_c) \, H(\gamma'; \gamma'_{\rm min}, \gamma'_b) +
```
cosimoNigro, line 575: should be `gamma'_c`, the class `ExpCutoffPowerLaw` has the same small mistake

agnpy/spectra/spectra.py
```diff
@@ -565,6 +566,131 @@ def __str__(self):
...
+            self.gamma_c,
+            self.gamma_b,
+            self.gamma_min,
+            self.gamma_max
+        )
+
+    def __str__(self):
+        return (
+            f"* {self.particle} energy distribution\n"
+            + f" - broken power law\n"
```
cosimoNigro, line 683: it's not a broken power law, please change the string with the name

agnpy/synchrotron/proton_synchrotron.py
```diff
@@ -0,0 +1,183 @@
+# module containing the synchrotron radiative process
+import numpy as np
+import astropy.units as u
+from astropy.constants import e, h, c, m_e, m_p, sigma_T, G
```
cosimoNigro, line 4: h, m_e, sigma_T and G are not needed.
If you use an IDE as visual studio, you can actually see which names (imports or defined variables) are not needed in the code.

agnpy/synchrotron/proton_synchrotron.py
```diff
@@ -0,0 +1,183 @@
...
+from ..utils.math import axes_reshaper, gamma_e_to_integrate
+from ..utils.conversion import nu_to_epsilon_prime, B_to_cgs, lambda_c_p
+from .synchrotron import Synchrotron
```
cosimoNigro, line 7: Also `Synchrotron` is not used.

agnpy/synchrotron/proton_synchrotron.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, blob, ssa=False, integrator=np.trapz):
+        self.blob = blob
+        self.ssa = ssa
+        self.integrator = integrator
+
+    @staticmethod
+    def evaluate_tau_ssa(
+        nu,
+        z,
+        d_L,
```
cosimoNigro, line 41: `d_L` is not used in this function

agnpy/synchrotron/synchrotron.py
```diff
@@ -1,9 +1,9 @@
 # module containing the synchrotron radiative process
 import numpy as np
 import astropy.units as u
-from astropy.constants import e, h, c, m_e, sigma_T
+from astropy.constants import e, h, c, m_e, m_p, sigma_T
```
cosimoNigro, line 4: `m_p` not used

agnpy/synchrotron/synchrotron.py
```diff
@@ -51,11 +51,11 @@ def epsilon_B(B):
     return (B / B_cr).to_value("")
 
 
-def single_electron_synch_power(B_cgs, epsilon, gamma):
+def single_particle_synch_power(B_cgs, epsilon, gamma, mass=m_e):
     """angle-averaged synchrotron power for a single electron,
```
cosimoNigro, line 55: for a single electron -> for a single particle of mass `mass`

agnpy/tests/test_proton_synchrotron.py
```diff
@@ -0,0 +1,140 @@
...
+R = 1e16 * u.cm #radius of the blob
+
+
+class TestProtonSynchrotron:
+    """Class grouping all tests related to the Proton Synchrotron class."""
+
+    def test_synch_reference_sed(self):
+        """Test agnpy proton synchrotron SED against the SED produced by the same particle distribution from Matteo Cerruti's code."""
+        # reference SED
+        lognu, lognuFnu = np.genfromtxt(f"{data_dir}/reference_seds/cerruti_psynch/test_pss.dat",  dtype = 'float', comments = '#', usecols = (0,4), unpack = True)
```
cosimoNigro, line 42: There is a function `extract_columns_sample_file` in `tests.utils`, used in all the other tests (e.g. see `test_synchrotron.py`)

cosimoNigro: Ah, ok, I see, the file has several columns and they are actually log values. SO maybe the function is not so useful in this case.

agnpy/tests/test_spectra.py
```diff
@@ -538,10 +538,127 @@ def test_init(self, p, gamma_c, gamma_min, gamma_max):
...
+        condition = (ebpwl.gamma_min <= gamma) * (gamma <= ebpwl.gamma_max)
+        # check that outside the boundaries values are all 0
+        assert not np.all(values[~condition])
+
+    @pytest.mark.parametrize("p1", np.arange(1, 2, 0.5))
+    @pytest.mark.parametrize("p2", np.arange(3, 4, 0.5))
+    @pytest.mark.parametrize("gamma_b", np.logspace(3, 5, 3))
+    @pytest.mark.parametrize("gamma_min", np.logspace(0, 2, 3))
+    @pytest.mark.parametrize("gamma_max", np.logspace(6, 8, 3))
+    @pytest.mark.parametrize("gamma_c", np.logspace(3, 5, 3))
```
cosimoNigro, line 558: I am just wondering what is the effect of having `gamma_c` < `gamma_b`, that the spectrum looks like a simple cutoff powerlaw?

agnpy/utils/conversion.py
cosimoNigro: I get your point @jsitarek, but if I am confused by this:
In the synchrotron SED formula there is actually a multiplication by epsilon (A.30 of [our paper](https://www.aanda.org/articles/aa/pdf/2022/04/aa42000-21.pdf)).
@dimaniad6 got a correct SED comparison with Cerruti by multiplying by an epsilon scaled to the proton rest mass energy. How would we obtain the same number if we were to have an epsilon only scaled by the electron mass? We should multiply by the ratio between the two masses, right?

If so, I am even more confused because I remembered there was an inverse cube mass in the synchrotron energy loss. Maybe it is absorbed in the R(x) function and in the epsilon factor?

cosimoNigro (changes requested):

cosimoNigro:
Thanks a lot for your work @dimaniad6!

Some more things, technicals and theoreticals

**Technical comments**:
1) Please, the next time that you implement changes to different classes, try to split them in different PRs. Otherwise PRs become very large and difficult to review. For example, the `ExpCutoffBrokenPowerLaw` implementation could have gone in a separate PR - though I understand in this particular case you implemented it just to validate the proton synchrotoron SED you got from Cerruti;

2) despite already merging one of your a PRs, you don't appear [among the contributors](https://github.com/cosimoNigro/agnpy/graphs/contributors). Also, in this PR, the icon of your profile appears in the first comment, but not alongside the individual commits (as it should be). [Here](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration) you can see how to correctly set your username and email. This should be sufficient for your contributions to be correctly identified and tagged. You can check your git credentials with `git config --list`; I suspect the problem might be that the mail is not correctly set.

**Theoretical comments**:
1) Is SSA relevant for proton synchrotron?
I copy-paste part of the content of a mail Julian sent me, I think it perfectly clarifies what are our doubts on SSA for proton synchrotron:
   
    > I am not sure if it is done correctly the same way for protons as for electrons. Moreover if we have both protons and electrons in the blob, electrons might be more efficient at the absorption of proton-synchrotron radiation. I do not
remember ever seeing SSA of proton synchrotron being considered in the papers, and doing a quick search I cannot find anything either. In normal situation I think it is negligible, because protons couple to photons much weaker than electrons do. I would suggest to simply drop the SSA part for the proton synchrotron.

    Since you introduced it just to reproduce Cerruti's SED, could you ask him to provide you with an SED without SSA? I think it is wise to drop it, if we are not 100% sure it matters and cannot find any other reference in the literature;

2) about the epsilon equivalencies with `m_p`, I have already commented in the code.
@jsitarek, I don't get very well your concern. Users will just plot $\nu$ vs $\nu F_{nu}$, and in this case the x axis is independent on the mass, right? Or are you concerned someone will plot using epsilon?

3) Our psynch SEDs are - across all the energy range - a factor 0.8 those of Matteo. Could you find out why?
proton_synch_comparison_PKS

agnpy/utils/conversion.py
jsitarek: Hi @cosimoNigro 
looking again at my comment, I think I wrote something wrong. Indeed plotting it vs nu is perfectly fine with e- and p together, plotting vs epsilon would be not correct, but as you wrote this is not something that users would do. In the total energy loss for synchrotron radiation of electrons you have sigma_T, which goes ~ 1/ m_e^2 
the differential emission is more complicated, because it includes the bessel function integration with limit being nu/nu_c, where nu_c ~ 1/m_e
see Eq 4.31 and 4.32 in Blumenthal & Gould 1970: 
https://journals.aps.org/rmp/abstract/10.1103/RevModPhys.42.237
(the formula is for emitted spectral power P(nu), which I guess should be EdN / dnu.
there is a formula also for low frequency case (4.34) with P(nu) ~ [1/m_e](nu/2nu_c)^1/3, which would end up as ~m^2/3

jsitarek:
>        Since you introduced it just to reproduce Cerruti's SED, could you ask him to provide you with an SED without SSA? I think it is wise to drop it, if we are not 100% sure it matters and cannot find any other reference in the literature;
> 
I'm pretty sure that in the previous post plot the SSA will not change anything at least > 10^10 Hz. 
The drop-off at lower frequency I'm not sure if it comes from SSA, or from the gamma_min of the proton distribution. 
Either way, even before asking Matteo to rerun it without SSA I would propose to just rerun the new agnpy code without SSA and see if it still matches. If needed we can limit the range to >10^10 Hz.

>     2. about the epsilon equivalencies with `m_p`, I have already commented in the code.
>        @jsitarek, I don't get very well your concern. Users will just plot ν vs νFnu, and in this case the x axis is independent on the mass, right? Or are you concerned someone will plot using epsilon?
> 
I think this was just a mistake on my side, sorry, it should be fine as it is . 

>     3. Our psynch SEDs are - across all the energy range - a factor 0.8 those of Matteo. Could you find out why?
>        proton_synch_comparison_PKS

as a quick test you could try to apply finer integration, but probably it will not help. 
otherwise to investigate it, it would be good to have a few more curves from Matteo, changing one parameter at the time: 
1) setting z to very small number: 0.01 to avoid problems with differnet cosmology used
2) checking if the ratio is the same for different doppler beaming, in particular you can try D=1.01 (with Gamma=1.004 or something like this), but making the blob non-relativistic might cause different troubles as well. 
3) running it for a different index of protons and seeing if the ratio keeps the same - this will tell you if the problem is somewhere in the general normalization or in the integrated parts.

agnpy/tests/test_spectra.py
dimaniad6: Yes, from the point where gamma_c = gamma_b, they look alike but they're not exactly the same (even if you put the normalization constant of the exp cut off equal to the one of exp bkn cut off times the gamma_b to the power of p) . They are, thought, for very high values of gamma

cosimoNigro:
Thanks a lot @IlariaViale, @dimaniad6, this is a fantastic new feature of the code!

Can one of you open an issue about the comparison with Cerruti, please?

PR #133, opened 2023-01-10
Absorption with cubepy

pawel21:
Hi,

I have created PR for my modification of code for absorption using cubepy.
I added example notebook in _agnpy/docs/tutorials_

### DIFF
--- a/agnpy/absorption/absorption.py
+++ b/agnpy/absorption/absorption.py
@@ -610,7 +610,7 @@ class Absorption:
             )
         return tau_blr_list
 
-    def tau_blr_lines_cubepy(self, nu, lines_list, eps_abs=1e-6):
+    def tau_blr_lines_cubepy(self, nu, lines_list=lines_dictionary.keys(), eps_abs=1e-6):
         """
         Calculate the absorption in all available BLR (Broad Line Region) shells.
         This function scales the radius and luminosity to the Hβ (H-beta) line


pawel21:
@cosimoNigro @jsitarek could you look on my PR? :)

cosimoNigro:
I will @pawel21, sorry for the delay. I had little time in general and an issue to fix before this PR.
Will try in the next two weeks.

docs/tutorials/absorption_in_BLR_with_cubepy.ipynb
```diff
@@ -0,0 +1,175 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "SUM_RATIO_LINES_BLR = 30.652\n",
+    "\n",
+    "def get_absor_blrs(E, L_disk, R_line, r_blob_bh, z):\n",
```
jsitarek, line 25: this function would be useful to be included in the main code of agnpy
XI_LINE and eps_abs  should be then among the function parameters

jsitarek (changes requested):

jsitarek:
sorry for a very late feedback on this
@pawel21 - I left a few comments on the PR
@cosimoNigro - how do you prefer to proceed - should this function substitute the current integration? or should it be added as a separate function (like Paweł did now)? In the latter case a test comparing the two methods should be added.
The former case might be more reasonable, because this method is more precise and do not have those crazy memore requirements of the standard integration scheme.

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,99 @@ def evaluate_tau_blr_mu_s(
...
+        nu,
+        eps_abs,
+        z,
+        mu_s,
+        L_disk,
+        xi_line,
+        epsilon_line,
+        R_line,
+        r,
+        l_size=50,
```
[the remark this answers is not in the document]
pawel21: I removed

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,99 @@ def evaluate_tau_blr_mu_s(
...
+            Luminosity of the disk whose radiation is being reprocessed by the BLR
+        xi_line : float
+            fraction of the disk radiation reprocessed by the BLR
+        epsilon_line : string
+            dimensionless energy of the emitted line
+        R_line : :class:`~astropy.units.Quantity`
+            radius of the BLR spherical shell
+        r : :class:`~astropy.units.Quantity`
+            distance between the Broad Line Region and the blob
+        l_size : int
```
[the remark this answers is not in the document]
pawel21: Done

jsitarek:
digging through my old e-mails I realized that this PR is still pending.
@pawel21 I think that still 2 of my comments are not implemented/responded
@cosimoNigro any feedback about the general implementation and relation to the older code?

cosimoNigro:
Sorry for letting it slip, will do the review later today.

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,96 @@ def evaluate_tau_blr_mu_s(
...
+        nu,
+        eps_abs,
+        z,
+        mu_s,
+        L_disk,
+        xi_line,
+        epsilon_line,
+        R_line,
+        r,
+        mu=mu_to_integrate,
```
cosimoNigro, line 459: the `mu` and `phi` arguments are not used in the function, you use directly the default `mu_to_integrate` and `phi_to_integrate`

agnpy/absorption/absorption.py
```diff
@@ -479,6 +571,62 @@ def tau_blr_mu_s(self, nu):
...
         )
 
+    def tau_blr_cubepy(self, nu, eps_abs=1e-6):
+        """Evaluates the gamma-gamma absorption produced by a spherical shell
+        BLR for a general set of model parameters with cubepy integration
+        method
+        """
+
+        tau_blr_list = []
+        for freq in nu:
```
cosimoNigro, line 581: I think the problem of using integrators as `cubepy`, that rely on functions, instead of `trapz` is that you do not get directly the output as an array but you are forced to loop over each frequency value. Anyhow, I think this is good for the moment.

agnpy/absorption/absorption.py
```diff
@@ -479,6 +571,62 @@ def tau_blr_mu_s(self, nu):
...
+            Absolute precision for the calculation, default is 1e-6
+        """
+        SUM_RATIO_LINES_BLR = 30.652
+        XI_TARGET_LINE = self.target.xi_line
+        tau_z_all = np.zeros(nu.shape)
+
+        blr_shells = dict()
+        absorption_shells = dict()
+
+        for line_key in lines_dictionary.keys():
```
cosimoNigro, line 617: I am a bit lost with the calculations in this `for` loop.
Are you assuming that the `target` is a `SphericalShellBLR` emitting the `Hbeta` line? If this is not the case then I have troubles in understanding why you scale radius and luminosity to the current (arbitrary) line.

cosimoNigro: If, on the other hand, the user has to use the `Hbeta` line as a target, then this should be stated explicitly somewhere.

cosimoNigro (commented):
Thanks a lot for your work @pawel21, and apologies again if I got disconnected and took a while to review it.

I think once you fix mine and @jsitarek's comments we can merge it in.
Discussing how to handle the general approach towards integration (e.g. allowing to choose between trapezoidal and adaptive integration) is something we should discuss together at a certain point in the future, along with the re-organisation of the package, not in this PR.

docs/tutorials/absorption_in_BLR_with_cubepy.ipynb
```diff
@@ -0,0 +1,275 @@
+{
```
cosimoNigro, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    SUM_RATIO_LINES_BLR = 30.652
</pre><p>If this notebook has to be included in the docs I think it needs to be considerably improved, at the moment it has no commentary and it does not explain what its purpose is.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cosimoNigro/agnpy/pull/133/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

cosimoNigro:
Another thing that is missing is an `absorption` function returning the actual attenuation factor in the `cubepy` case, now only `tau` is computed.

agnpy/absorption/absorption.py
pawel21: Done

agnpy/absorption/absorption.py
pawel21: Hbeta is used as a reference line for radius and luminosity scaling to ensure consistency in calculations for different BLR lines.

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,96 @@ def evaluate_tau_blr_mu_s(
...
+
+        Returns
+        -------
+        :class:`~astropy.units.Quantity`
+            array of the tau values corresponding to each frequency
+        """
+        # conversions
+        epsilon_1 = nu_to_epsilon_prime(nu, z)
+
+        def f(x):
```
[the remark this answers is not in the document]
pawel21: Done

agnpy/absorption/absorption.py
jsitarek: true, but since you do not have arrays you have much smaller memory usage, and the integration module automatically selects the points to evaluate the function in an optimal way, so a simple loop over frequencies is a small price to pay :-)

pgliwny:
> Another thing that is missing is an `absorption` function returning the actual attenuation factor in the `cubepy` case, now only `tau` is computed.

What do you mean? Which function is missing? I tried to add a function similar to a function for another target.

docs/tutorials/absorption_in_BLR_with_cubepy.ipynb
pgliwny: Let's remove the notebook for now so we can go further. I would also like to develop other features, but it will be easier if the code for calculation absorption with cubepy is in the main repo. In the next step, I will add notebooks with explanations for the multi-shell BLR model.

pgliwny:
@jsitarek  @cosimoNigro  Could you please review my code? 
I have addressed all of your suggestions and modified the code accordingly. I will be using this code to model absorption for new source in near future and to add a fit that allows for the consideration of multiple shells included in the BLR. For this purpose, it would be easier to have this code in the main repository for further improvements.

jsitarek:
Hi @pgliwny 
The automatic CI checks fail:

FAILED agnpy/fit/tests/test_fit.py::TestFit::test_mrk_421_fit - IndexError: boolean index did not match indexed array along dimension 2; dimension is 1 but corresponding boolean dimension is 9
FAILED agnpy/fit/tests/test_fit.py::TestFit::test_pks_1510_fit - ValueError: non-broadcastable output operand with shape (6,1,1) doesn't match the broadcast shape (6,1,6)

maybe your PR changed something in the API

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,113 @@ def evaluate_tau_blr_mu_s(
...
+            Cosine of the angle between the blob motion and the jet axis.
+        epsilon_1 : float
+            Dimensionless energy parameter for the incoming photon.
+        epsilon_line : float
+            Dimensionless energy of the emitted line.
+
+        Returns
+        -------
+        float
+            The value of the integrand for the given parameters.
```
jsitarek, line 470: it would be useful to specify in which units the integrand is specified

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,113 @@ def evaluate_tau_blr_mu_s(
...
+        """
+        mu = cubepy_params_array[0]
```
jsitarek, line 472: ```(mu, phi, log_l) = cubepy_params_array``` should work the same and would be more concise

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,113 @@ def evaluate_tau_blr_mu_s(
         )
         return (prefactor * integral).to_value("")
 
+    @staticmethod
+    def compute_integrand(cubepy_params_array, R_line, mu_s, epsilon_1, epsilon_line):
```
jsitarek, line 449: since the function is only for BLR and it is in general Absorption class it would be good to have 'blr' in the name

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,113 @@ def evaluate_tau_blr_mu_s(
...
+        R_line : :class:`~astropy.units.Quantity`
+            radius of the BLR spherical shell
+        r : :class:`~astropy.units.Quantity`
+            distance between the Broad Line Region and the blob
+        mu, phi : :class:`~numpy.ndarray`
+            arrays of cosine of zenith and azimuth angles to integrate over
+
+        Returns
+        -------
+        :class:`~astropy.units.Quantity`
```
jsitarek, line 521: tau is dimensionless, and you return ..to_value (""), so I think it should be simply a float

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,113 @@ def evaluate_tau_blr_mu_s(
...
+        BLR for a general set of model parameters using cubepy
+
+        Parameters
+        ----------
+        nu : :class:`~astropy.units.Quantity`
+            array of frequencies, in Hz, to compute the tau
+            **note** these are observed frequencies (observer frame)
+        z : float
+            redshift of the source
+        mu_s : float
```
jsitarek, line 504: I think this function is following evaluate_tau_blr that is only valid for mu_s=1 (otherwise integration over path of the photon and over the distance from the black hole is not equivalent), so it should not have it as a parameter  (@cosimoNigro can you also have a look at this?)

agnpy/absorption/absorption.py
```diff
@@ -479,6 +588,67 @@ def tau_blr_mu_s(self, nu):
...
+        their respective ratios.
+
+        Parameters
+        ----------
+        nu: numpy array
+            Frequency array with unit
+        eps_abs: float, optional
+            Absolute precision for the calculation, default is 1e-6
+        """
+        SUM_RATIO_LINES_BLR = 30.652
```
jsitarek, line 628: hardcoding this is not safe, you should sum up the lines from the dictionary that you will use

agnpy/absorption/absorption.py
jsitarek: I think this is solved already (the explanation is in the docstring of the function), resolving..

agnpy/absorption/absorption.py
```diff
@@ -479,6 +588,67 @@ def tau_blr_mu_s(self, nu):
...
+        XI_TARGET_LINE = self.target.xi_line
+        tau_z_all = np.zeros(nu.shape)
+
+        blr_shells = dict()
```
jsitarek, line 632: why do you define those dictionaries? you are using them only to sum up over all the lines, so you could instead created SphericalShellBLR and Absorption as temporary variables directly in the for loop

agnpy/absorption/absorption.py
```diff
@@ -479,6 +588,67 @@ def tau_blr_mu_s(self, nu):
...
+        return tau_blr_list
+
+    def tau_blr_all_lines_cubepy(self, nu, eps_abs=1e-6):
+        """
+        Calculate the absorption in all available BLR (Broad Line Region) shells.
+        This function scales the radius and luminosity to the Hbeta line
+        and applies the computation to all the available BLR lines (shells) using
+        their respective ratios.
+
+        Parameters
```
jsitarek, line 621: how about adding a dictionary 'lines' to the parameters (and setting it by default to lines_dictionary)
this would allow to run the function with limited number of lines (selecting only the most important ones for speed) or with modified dictionary (if we have more accurate measurements for a particular source). It would also facilitate tests because you could then implement a quick test on a dictionary of 2 lines

agnpy/absorption/tests/test_absorption.py
```diff
@@ -228,6 +228,31 @@ def test_abs_dt_vs_point_source(self):
         # requires a 10% deviation from the two SED points
         assert check_deviation(nu, tau_dt, tau_ps_dt, 0.1)
 
+    def test_abs_blr_cubepy(self):
```
jsitarek, line 231: the test as it is written now does not test much: only if you do not have a crash because of not matching API, or some internal inconsistency in the code. 
Instead you could run a test of comparing tau_blr_cubepy output with tau_blr

jsitarek (changes requested):
I left some comments. In particular the test should be rewritten.

agnpy/absorption/absorption.py
pgliwny: renamed to `compute_integrand_for_blr`

agnpy/absorption/absorption.py
pgliwny: done

agnpy/absorption/absorption.py
pgliwny: Great idea, but what do you think about doing it in the "second" iteration (after merging this code with all lines into the main branch)?

jsitarek: if you prefer ok, but note that this is really minor change to the code, just adding a new parameter to the function, with a default value and then in the line 637 using this parameter instead of lines_dictionary.
And it would simplify implementation of my comment about tests

agnpy/absorption/absorption.py
pgliwny: Done

agnpy/absorption/absorption.py
pgliwny: I have created a new function, `tau_blr_lines_cubepy`. Could you look?

agnpy/absorption/absorption.py
jsitarek: resolving an old comment

agnpy/absorption/absorption.py
```diff
@@ -479,6 +586,116 @@ def tau_blr_mu_s(self, nu):
...
+            absorption_shells[line_key] = Absorption(blr_shells[line_key], r=self.r, z=self.z)
+
+            tau_z_current = absorption_shells[line_key].tau_blr_cubepy(nu, eps_abs=eps_abs)
+            tau_z_all += tau_z_current
+
+        return tau_z_all
+
+    def tau_blr_lines_cubepy(self, nu, lines_list, eps_abs=1e-6):
+        """
+        Calculate the absorption in all available BLR (Broad Line Region) shells.
```
jsitarek, line 652: "all available BLR shells" is a copy paste from the original function, better write "multiple BLR ..." or something like this since now you specify exaclty the list of lines to consider

agnpy/absorption/absorption.py
```diff
@@ -479,6 +586,116 @@ def tau_blr_mu_s(self, nu):
...
+        eps_abs : float, optional
+            Absolute precision for the calculation (default is 1e-6).
+
+        Returns
+        -------
+        tau : numpy.ndarray
+             Optical depth values corresponding to the input frequencies.
+        """
+
+        sum_ratio_lines = sum(lines_dictionary[line]['L_Hbeta_ratio'] for line in lines_list)
```
jsitarek, line 675: you need to introduce a check or even better throw an exception if any of the requested line is not in lines_dictionary.keys()

agnpy/absorption/absorption.py
```diff
@@ -479,6 +586,116 @@ def tau_blr_mu_s(self, nu):
...
+        Calculate the absorption in all available BLR (Broad Line Region) shells.
+        This function scales the radius and luminosity to the Hβ (H-beta) line
+        and computes the absorption for all available BLR lines (shells) using
+        their respective ratios.
+
+        Parameters
+        ----------
+        nu : numpy.ndarray
+            Frequency array for which the absorption will be calculated.
+        lines_list : list of str
```
jsitarek, line 661: if you set this as default to lines_dictionary.keys() then you do not need `tau_blr_all_lines_cubepy` function any more

agnpy/absorption/absorption.py
jsitarek: thanks. I put some comments there, and I am closing the comment here

agnpy/absorption/absorption.py
```diff
@@ -479,6 +586,116 @@ def tau_blr_mu_s(self, nu):
...
+        -------
+        tau : numpy.ndarray
+             Optical depth values corresponding to the input frequencies.
+        """
+
+        sum_ratio_lines = sum(lines_dictionary[line]['L_Hbeta_ratio'] for line in lines_list)
+        XI_TARGET_LINE = self.target.xi_line
+        tau_z_lines = np.zeros(nu.shape)
+
+        blr_shells = dict()
```
jsitarek, line 679: copying the comment from the previous function:



why do you define those dictionaries? I guess it is a remnant of a script where you were also plotting the effect of individual lines, but this is either way not possible this function.

here you are using the dictionaries only to sum up over all the lines, so you could instead created SphericalShellBLR and Absorption as temporary variables directly in the for loop

agnpy/absorption/absorption.py
pawel21: Done

agnpy/absorption/tests/test_absorption.py
pawel21: Done, added `test_tau_accuracy_standard_vs_cubepy`

agnpy/absorption/absorption.py
pawel21: removed this function for all lines

agnpy/absorption/absorption.py
pawel21: removed

agnpy/absorption/absorption.py
pawel21: done

agnpy/absorption/absorption.py
pawel21: `all_lines` function removed

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,112 @@ def evaluate_tau_blr_mu_s(
...
+            Luminosity of the disk whose radiation is being reprocessed by the BLR
+        xi_line : float
+            fraction of the disk radiation reprocessed by the BLR
+        epsilon_line : string
+            dimensionless energy of the emitted line
+        R_line : :class:`~astropy.units.Quantity`
+            radius of the BLR spherical shell
+        r : :class:`~astropy.units.Quantity`
+            distance between the Broad Line Region and the blob
+        mu, phi : :class:`~numpy.ndarray`
```
jsitarek, line 515: those two are not in the list of parameters

agnpy/absorption/absorption.py
```diff
@@ -443,6 +445,112 @@ def evaluate_tau_blr_mu_s(
...
+        -------
+        :class:float
+            array of the tau values corresponding to each frequency
+        """
+        # conversions
+        epsilon_1 = nu_to_epsilon_prime(nu, z)
+        # set boundary for integration
+        low = np.array(
+            [
+                [min(mu_to_integrate)],
```
jsitarek, line 528: the integration of mu and phi should be simply -1 to 1 and 0 to 2 pi.

agnpy/absorption/tests/test_absorption.py
```diff
@@ -228,6 +228,49 @@ def test_abs_dt_vs_point_source(self):
...
```
jsitarek, line 231: now that you added test_tau_accuracy_standard_vs_cubepy this test does not seem to add anything to it, so I think you can remove test_abs_blr_cubepy

PR #145, opened 2023-05-16
Fix the issues with the InterpolatedDistributions

### LINKED ISSUE #135: Error in InterpolatedDistribution
Hello, when I use InterpolatedDistribution, there was a problem

6ADD06A7-F45B-4495-B56C-4733A3BDC02D

cosimoNigro:
Hello,

this PR fixes the issues reported in #135 with the `InterpolatedDistribution`.

The latter could not be used with the classes corresponding to the radiative processes as it had no parameter.
I added a `norm` parameter that is just a multiplicative factor of the interpolated distribution.
The latter is propagated to the gammapy and sherpa wrappers as a `log10_norm` fit parameter.

Fixes #135.


cosimoNigro:
Will try to extend the code coverage. I added some checks in the wrappers, in case the `InterpolatedDistribution` is used.

cosimoNigro:
These last commits should finish fixing the usage of `InterpolatedDistribution` by the wrappers and also increase the test coverage.

PR #149, opened 2023-07-05
Start to implement photo-meson interactions

cosimoNigro:
@IlariaViale, @dimaniad6, let us use this branch to work on the implementation of the photomeson interactions.
We are following the analytical parametrisation of Kelner Aharonian 2008 (https://arxiv.org/abs/0803.0688).


cosimoNigro:
Hello @dimaniad6, @IlariaViale, do you think you will have some time in the next days to work on this?

IlariaViale:
Hi Cosimo,

I think Dimitri is having some oral examinations for the phd applications
in these days.
In this month I mainly focused on starting writing the thesis so I did not
proceed that much. I have some preliminary results which I obtained ~1
month ago: the spectrum is similar to that of Aharonian but not exactly the
same.
If you want in the next days I can show you what I did, but Dimitri was
also interested in this so I would wait for him to be available.
In any case, yes, in the next weeks I will try to share my time on both
writing and the code in order to do both.

Ilaria

Il giorno ven 1 set 2023 alle ore 11:19 Cosimo Nigro <
***@***.***> ha scritto:

dimaniad6:
Hello Cosimo and Ilaria,

I am sorry for the late reply, I just came back from vacations in Anafi
where having a signal was quite complicated.

My schedule this period is that I have an interview for Napoli in the 11th
of September, for which I have to study a lot,
plus probably for Palermo. I also have to finish the application for
Barcelona and prepare for SIF (I think I will use the same presentation
as the one I did on IFAE). After SIF (12-15 of sept) I will go to Padova to
take some stuff and see some people and return back on the 22th of
September.
After that I will have plenty of time to work and finally finish the
implementation of p-gamma with Ilaria.

Cheers,
Dimitris

cosimoNigro:
Thanks for the updates @dimaniad6, @IlariaViale, I did not mean to rush you.

By the way, when you reply to GitHub notifications via mail your reply is also posted on GitHub 😄, I am removing your replies since they contain personal informations.

PR #156, opened 2024-03-26
spectra documentation fix

grzegorzbor:
Hi Cosimo,

I think i found a small bug in the documentation, here is the fix. Can you take a look?


jsitarek (approved):
the change looks fine to me

cosimoNigro:
Thank you @grzegorzbor

PR #157, opened 2024-04-20
Synchrotron losses time evolution

grzegorzbor:
The ParticleDistribution.evaluate_time() function added.
It's a basic method that accepts the energy losses function as the parameter, and evaluates it for the requested time period (with optional subperiods). 
For now  just the first, basic version. It can be later extended to be smarter - for example automatically use shorter time ranges for the energies where higher losses occur.
For now only synchrotron losses implemented, other will be added later.
We can later add the higher level function that would take the flags which radiative processes should be used, instead of accepting generic energy loss function as the parameter.

### DIFF
--- a/agnpy/spectra/spectra.py
+++ b/agnpy/spectra/spectra.py
@@ -1,11 +1,10 @@
 # module containing the particle spectra
 import numpy as np
 import astropy.units as u
-from astropy.units.quantity import Quantity
+from agnpy.utils.conversion import mec2
 from astropy.constants import m_e, m_p, c
 from scipy.interpolate import CubicSpline
 import matplotlib.pyplot as plt
-import math
 
 
 __all__ = [
@@ -201,16 +200,34 @@ class ParticleDistribution:
         return ax
 
     def evaluate_time(self, time, energy_loss_function, subintervals_count=1):
+        """Performs the time evaluation of energy change for this particle distribution.
+
+        Parameters
+        ----------
+        time : `~astropy.units.Quantity`
+            total time for the calculation
+        energy_loss_function : function
+            thr function to be used for calculation of energy loss;
+            the function accepts the gamma value and produces "energy per time" quantity
+        subintervals_count : int
+            optional number defining how many equal-length subintervals the total time will be split into
+
+        Returns
+        -------
+        InterpolatedDistribution
+            a new distribution
+        """
         unit_time_interval = time / subintervals_count
 
         def gamma_recalculated_after_loss(gamma):
-            old_energy = (gamma * m_e * c ** 2).to("J")
-            energy_loss_per_time = energy_loss_function(gamma).to("J s-1")
-            energy_loss = (energy_loss_per_time * unit_time_interval).to("J")
+            old_energy = (gamma * mec2).to("erg")
+            energy_loss_per_time = energy_loss_function(gamma).to("erg s-1")
+            energy_loss = (energy_loss_per_time * unit_time_interval).to("erg")
             new_energy = old_energy - energy_loss
             if np.any(new_energy < 0):
-                raise ValueError("Energy loss formula returned value higher then original energy")
-            new_gamma = (new_energy / (m_e * c ** 2)).value
+                raise ValueError(
+                    "Energy loss formula returned value higher then original energy. Use the shorter time ranges.")
+            new_gamma = (new_energy / mec2).value
             return new_gamma
 
         gammas = self.gamma_values()
--- a/agnpy/spectra/spectra.py
+++ b/agnpy/spectra/spectra.py
@@ -199,7 +199,7 @@ class ParticleDistribution:
 
         return ax
 
-    def evaluate_time(self, time, energy_loss_function, subintervals_count=1):
+    def evaluate_time(self, time, energy_loss_function, subintervals_count=10):
         """Performs the time evaluation of energy change for this particle distribution.
 
         Parameters
@@ -226,11 +226,11 @@ class ParticleDistribution:
             new_energy = old_energy - energy_loss
             if np.any(new_energy < 0):
                 raise ValueError(
-                    "Energy loss formula returned value higher then original energy. Use the shorter time ranges.")
+                    "Energy loss formula returned value higher then original energy. Use shorter time ranges.")
             new_gamma = (new_energy / mec2).value
             return new_gamma
 
-        gammas = self.gamma_values()
+        gammas = np.logspace(np.log10(self.gamma_min), np.log10(self.gamma_max), 200)
         n_array = self.__call__(gammas)
 
         # for each gamma point create a narrow bin, calculate the energy loss for start and end of the bin,
@@ -243,12 +243,8 @@ class ParticleDistribution:
             narrowed_bins = bins_end_recalc - gammas
             if np.any(narrowed_bins <= 0):
                 raise ValueError(
-                    "Energy loss formula returned too big value - use shorter time intervals")
+                    "Energy loss formula returned too big value. Use shorter time ranges.")
             density_increase = bins_width / narrowed_bins
-            if np.any(density_increase < 1):
-                # not possible for simple scenarios because higher gammas should lose more energy,
-                # but maybe for more complex scenarios it could happen? then we should revise this assertion
-                raise AssertionError("Unexpected situation, bin has been widened instead of narrowed down")
             n_array = n_array * density_increase
 
         return InterpolatedDistribution(gammas, n_array)
@@ -304,9 +300,6 @@ class PowerLaw(ParticleDistribution):
 
     def __call__(self, gamma):
         return self.evaluate(gamma, self.k, self.p, self.gamma_min, self.gamma_max)
-
-    def gamma_values(self):
-        return np.logspace(np.log10(self.gamma_min), np.log10(self.gamma_max), 200)
 
     @staticmethod
     def evaluate_SSA_integrand(gamma, k, p, gamma_min, gamma_max):
@@ -787,7 +780,7 @@ class InterpolatedDistribution(ParticleDistribution):
         function to be used for integration, default is :class:`~numpy.trapz`
     """
 
-    def __init__(self, gamma, n, norm=1, mass=m_e, integrator=np.trapz, gamma_min=None, gamma_max=None):
+    def __init__(self, gamma, n, norm=1, mass=m_e, integrator=np.trapz):
         super().__init__(mass, integrator, "InterpolatedDistribution")
         if n.unit != u.Unit("cm-3"):
             raise ValueError(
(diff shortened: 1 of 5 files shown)


agnpy/spectra/spectra.py
```diff
@@ -1,9 +1,11 @@
 # module containing the particle spectra
 import numpy as np
 import astropy.units as u
-from astropy.constants import m_e, m_p
+from astropy.units.quantity import Quantity
+from astropy.constants import m_e, m_p, c
 from scipy.interpolate import CubicSpline
 import matplotlib.pyplot as plt
+import math
```
jsitarek, line 8: do you use this one?

agnpy/spectra/spectra.py
```diff
@@ -198,6 +200,42 @@ def plot(self, gamma=None, gamma_power=0, ax=None, **kwargs):
 
         return ax
 
+    def evaluate_time(self, time, energy_loss_function, subintervals_count=1):
```
jsitarek, line 203: the function should have a docstring explaining its parameters

jsitarek, line 203: I would set by default the subintervals_count to a higher number, maybe 10

agnpy/spectra/spectra.py
```diff
@@ -198,6 +200,42 @@ def plot(self, gamma=None, gamma_power=0, ax=None, **kwargs):
...
+        unit_time_interval = time / subintervals_count
+
+        def gamma_recalculated_after_loss(gamma):
+            old_energy = (gamma * m_e * c ** 2).to("J")
```
jsitarek, line 207: most of astrophysical codes is using CGS, so better to convert here and in the next lines to erg, erg / s, etc.

jsitarek, line 207: m_e c^2 is already defined in agnpy: https://github.com/cosimoNigro/agnpy/blob/75ab0b7fc25af8f8234b49cf6c33da0888baec7e/agnpy/utils/conversion.py#L6

agnpy/spectra/spectra.py
```diff
@@ -198,6 +200,42 @@ def plot(self, gamma=None, gamma_power=0, ax=None, **kwargs):
...
+            energy_loss_per_time = energy_loss_function(gamma).to("J s-1")
+            energy_loss = (energy_loss_per_time * unit_time_interval).to("J")
+            new_energy = old_energy - energy_loss
+            if np.any(new_energy < 0):
+                raise ValueError("Energy loss formula returned value higher then original energy")
```
jsitarek, line 212: you can add in the error message a suggestion to use more fine binning of time

agnpy/spectra/spectra.py
```diff
@@ -198,6 +200,42 @@ def plot(self, gamma=None, gamma_power=0, ax=None, **kwargs):
...
+            new_gamma = (new_energy / (m_e * c ** 2)).value
```
jsitarek, line 213: mec2

agnpy/spectra/spectra.py
```diff
@@ -198,6 +200,42 @@ def plot(self, gamma=None, gamma_power=0, ax=None, **kwargs):
...
+            bin_size_factor = 0.0001
+            bins_width = gammas * bin_size_factor
+            bins_end_recalc = gamma_recalculated_after_loss(gammas + bins_width)
+            gammas = gamma_recalculated_after_loss(gammas)
+            narrowed_bins = bins_end_recalc - gammas
+            if np.any(narrowed_bins <= 0):
+                raise ValueError(
+                    "Energy loss formula returned too big value - use shorter time intervals")
+            density_increase = bins_width / narrowed_bins
+            if np.any(density_increase < 1):
```
jsitarek, line 231: for synchrotron this is indeed true, but as the comment says other scenarios and e.g. increase of energy due to acceleration would not behave like this, so I would drop the check (it can be implemented in the tests however)

agnpy/spectra/spectra.py
```diff
@@ -250,6 +288,9 @@ def evaluate(gamma, k, p, gamma_min, gamma_max):
     def __call__(self, gamma):
         return self.evaluate(gamma, self.k, self.p, self.gamma_min, self.gamma_max)
 
+    def gamma_values(self):
+        return np.logspace(np.log10(self.gamma_min), np.log10(self.gamma_max), 200)
```
jsitarek, line 292: this 200 would be better as an optional parameter

agnpy/spectra/tests/test_spectra.py
```diff
@@ -740,3 +743,46 @@ def test_interpolation_physical(self):
...
         )
+
+
+class TestSpectraTimeEvolution:
+
+    @pytest.mark.parametrize("p", [0.5, 1.2, 2.4])
+    @pytest.mark.parametrize("time", [1, 60, 120] * u.s)
+    def test_compare_numerical_results_with_analytical_calculation(self, p, time):
+        """Test time evolution of spectral electron density for synchrotron energy losses.
+         Use a simple power low spectrum for easy calculation of analytical results."""
```
jsitarek, line 754: power law

agnpy/spectra/tests/test_spectra.py
```diff
@@ -740,3 +743,46 @@ def test_interpolation_physical(self):
...
+            """Reverse-time calculation of the gamma value before the synchrotron energy loss,
+             using formula -dE/dt ~ (E ** 2)"""
+            coef = synch._electron_energy_loss_formula_prefix() / (m_e * c ** 2)
+            return 1 / ((1 / gamma_after_time) - time * coef)
+
+        def integral_analytical(gamma_min, gamma_max):
+            """Integral for the power-law distribution"""
+            return k * (gamma_max ** (1 - p) - gamma_min ** (1 - p)) / (1 - p)
+
+        assert u.isclose(
```
jsitarek, line 773: I would add a comment here: the synchrotron energy losses are the largest for the highest energy, so we test here only the highest energy as the most affected

agnpy/spectra/tests/test_spectra.py
```diff
@@ -740,3 +743,46 @@ def test_interpolation_physical(self):
...
+            evaluated_n_e.integrate(evaluated_n_e.gamma_min, evaluated_n_e.gamma_max),
+            integral_analytical(gamma_before(evaluated_n_e.gamma_min, time), gamma_before(evaluated_n_e.gamma_max, time)),
+            rtol=0.05
+        )
+        assert u.isclose(
+            # only the highest energy range where most losses occur
+            evaluated_n_e.integrate(evaluated_n_e.gamma_max / 10, evaluated_n_e.gamma_max),
+            integral_analytical(gamma_before(evaluated_n_e.gamma_max / 10, time), gamma_before(evaluated_n_e.gamma_max, time)),
+            rtol=0.05
+        )
```
jsitarek, line 788: I think you can add another test function comparing if you get (nearly) the same if you:
1) iterate over e.g. 60 s in 10 steps
2) iterate over the same 60 s but in 20 steps
3) iterate first over 30 s, and then again (starting with interpolated distribution rather than power-law!)  over the remaining 30 s

agnpy/synchrotron/synchrotron.py
```diff
@@ -90,6 +90,18 @@ def __init__(self, blob, ssa=False, integrator=np.trapz):
         self.ssa = ssa
         self.integrator = integrator
 
+    def electron_energy_loss_per_time(self, gamma):
+        # For synchrotron, energy loss dE/dt formula is:
+        # (4/3 * sigma_T * c * magn_energy_dens) * gamma^2
+        # The part in brackets is fixed, so as an improvement we could calculate it once and cache,
```
jsitarek, line 96: ....is fixed for constant B...

agnpy/synchrotron/synchrotron.py
```diff
@@ -90,6 +90,18 @@ def __init__(self, blob, ssa=False, integrator=np.trapz):
...
+        # and later we could use the formula (fixed * gamma^2)
+        fixed = self._electron_energy_loss_formula_prefix()
+        return (fixed * gamma ** 2).to("erg/s")
+
+    def _electron_energy_loss_formula_prefix(self):
+        magn_energy_dens = (self.blob.B.to("T") ** 2) / (2 * mu0)
```
jsitarek, line 102: if you want to keep it in SI, it would be nice to add a comment that it is done like this because of the ambiginuity of the different CGS systems in terms of mu0 definition

jsitarek (changes requested):
thanks for the PR. 
I put some comments

agnpy/spectra/spectra.py
grzegorzbor: done

agnpy/spectra/tests/test_spectra.py
grzegorzbor: done

agnpy/spectra/spectra.py
grzegorzbor: removed

agnpy/spectra/spectra.py
grzegorzbor: done

agnpy/synchrotron/synchrotron.py
grzegorzbor: done

agnpy/spectra/spectra.py
grzegorzbor: done

agnpy/spectra/spectra.py
grzegorzbor: this function turned out not to be really needed, so i removed it

agnpy/spectra/tests/test_spectra.py
grzegorzbor: done, with slightly different, but similar, setup

agnpy/synchrotron/synchrotron.py
grzegorzbor: done

jsitarek (approved):
thanks for the changes, everything looks good to me now.
let's see what @cosimoNigro thinks

jsitarek:
then I think the simplest solution for the moment would be to fix in the environment file gammapy version to ~1.1.
Actually @cosimoNigro why gammapy is installed via pip and not directly?

grzegorzbor:
Hi @cosimoNigro 
I updated the dependencies of the project to the combination of the versions that seem to work fine. Also aligned the environment.yml and setup.py files to have the same dependency specification. Now all tests pass. 
Unfortunately the build is failing now on the next step, codecoverage, because the call to the codecov API returns an error "Rate limit reached. Please upload with the Codecov repository upload token to resolve issue". I tried a couple of times and it always ends with the same error, so it doesn't seem to be temporary. According to the codecov documentation https://docs.codecov.com/docs/quick-start you need to create an account on codecov, then generate a token, and then update the CI Test workflow definition with the token. For open source projects it should be free of charge.

cosimoNigro:
Thanks for the work @grzegorzbor and thanks @jsitarek for the review.
Everything looks good to me. I will take care of the failing test related to `Gammapy` and of the `codecov` issue.

PR #160, opened 2024-07-10
Photomesons interactions

cerasole:
First trial of implementation of Eqs. 69-70 in Kelner&Aharorian, plus a notebook to proceed with the validation.


cosimoNigro:
Thanks a lot @cerasole, I am merging this one in order to go ahead with the other pending PRs.
This merge should not create problem as it goes to `cosimoNigro/photomesons_interactions`.
We will continue the work from there, feel free to do any push onto that branch.
Thanks.

PR #163, opened 2024-10-08
Trying to fix issue in uploading to codecov

cosimoNigro:
This PR attempts to fix the issue in uploading to codecov according to @grzegorzbor suggestions in #157.


cosimoNigro:
Thanks @grzegorzbor for the tips.

PR #173, opened 2025-03-01
Time evolution implemented for Synchrotron Self Compton

grzegorzbor:
SSC time evolution implemented

### DIFF
--- a/agnpy/compton/synchrotron_self_compton.py
+++ b/agnpy/compton/synchrotron_self_compton.py
@@ -1,13 +1,14 @@
 # module containing the synchrotron self Compton radiative process
 import numpy as np
-from astropy.constants import c, sigma_T
-import astropy.units as u
+from astropy.constants import c, sigma_T, m_e
+
 from .kernels import isotropic_kernel
+from .. import Blob
 from ..synchrotron import Synchrotron
 from ..utils.math import (
     axes_reshaper,
     gamma_e_to_integrate,
-    nu_to_integrate,
+    nu_to_integrate, log
 )
 from ..utils.conversion import nu_to_epsilon_prime
 from ..radiative_process import RadiativeProcess
@@ -32,6 +33,42 @@ class SynchrotronSelfCompton(RadiativeProcess):
         self.blob = blob
         self.ssa = ssa
         self.integrator = integrator
+
+    def electron_energy_loss_rate_thomson(self, gamma):
+        return self._electron_energy_loss_thomson_formula_prefactor() * gamma ** 2
+
+    def _electron_energy_loss_thomson_formula_prefactor(self):
+        return ((4 / 3) * sigma_T * c * self.blob.u_ph_synch).to("erg/s")
+
+    def electron_energy_loss_rate(self, gamma):
+        epsilon = nu_to_epsilon_prime(nu_to_integrate, self.blob.z, self.blob.delta_D)
+        sed_flux_synch = Synchrotron.evaluate_sed_flux(
+            nu_to_integrate,
+            self.blob.z,
+            self.blob.d_L,
+            self.blob.delta_D,
+            self.blob.B,
+            self.blob.R_b,
+            self.blob.n_e,
+            *self.blob.n_e.parameters,
+            gamma=gamma_e_to_integrate
+        )
+
+        # differential energy density of photons
+        u_epsilon = self.blob.u_ph_synch_diff(epsilon, sed_flux_synch)
+
+        # reshape for multidimensional integration
+        _gamma, _epsilon = axes_reshaper(gamma, epsilon)
+
+        # formulas from paragraph 2.1, [Moderski2005]
+        b = 4 * _gamma * _epsilon # matrix [gamma x epsilon] - b < 1 means Thomson regime, b > 1 means K-N regime
+        # convert b value to the K-N correction parameter:
+        f_KN = np.where(b < 1e4, 1 / (1 + b) ** 1.5, (9 / (2 * b**2)) * (log(b) - 11/6))
+        # fold K-N correction parameters with the corresponding photon energy densities:
+        F_KN = np.trapz(f_KN * u_epsilon, epsilon)
+        # F_KN is basically total energy density corrected to KN regime for each gamma value
+        energyLossRate = (4 / 3) * sigma_T * c * F_KN * gamma ** 2
+        return energyLossRate.to("erg/s")
 
     @staticmethod
     def evaluate_sed_flux(
@@ -102,13 +139,7 @@ class SynchrotronSelfCompton(RadiativeProcess):
             integrator=integrator,
             gamma=gamma,
         )
-        # Eq. 8 [Finke2008]_
-        u_synch = (3 * np.power(d_L, 2) * sed_synch) / (
-            c * np.power(R_b, 2) * np.power(delta_D, 4) * epsilon
-        )
-        # factor 3 / 4 accounts for averaging in a sphere
-        # not included in Dermer and Finke's papers
-        u_synch *= 3 / 4
+        u_synch = Blob.evaluate_u_ph_synch_diff(R_b, d_L, delta_D, epsilon, sed_synch)
         # multidimensional integration
         _gamma, _epsilon, _epsilon_s = axes_reshaper(gamma, epsilon, epsilon_s)
         V_b = 4 / 3 * np.pi * np.power(R_b, 3)
--- a/agnpy/emission_regions/blob.py
+++ b/agnpy/emission_regions/blob.py
@@ -1,10 +1,14 @@
 """This module describes the emission regions responsible for the
 acceleration of particles to relativistic energies. Beside physical quantities
 related to the emission itself it contains the electrons energy distributions"""
+import numbers
+
 import numpy as np
 import astropy.units as u
 from astropy.coordinates import Distance
 from astropy.constants import c, sigma_T, m_e
+
+from .. import InterpolatedDistribution, ParticleDistribution
 from ..spectra import PowerLaw
 from ..utils.conversion import mec2, mpc2, B_to_cgs
 
@@ -60,13 +64,16 @@ class Blob:
         delta_D=10,
         Gamma=10,
         B=1 * u.G,
-        n_e=PowerLaw(mass=m_e),
-        n_p=None,
+        n_e : ParticleDistribution = PowerLaw(mass=m_e),
+        n_p : ParticleDistribution = None,
         xi=1.0,
         gamma_e_size=200,
         gamma_p_size=200,
         cosmology=None
     ):
+        if not isinstance(delta_D, numbers.Number) or delta_D <= 0:
+            raise ValueError("delta_D must be a positive number")
+
         self.R_b = R_b.to("cm")
         self.z = z
         # if the luminosity distance is not specified, it will be computed from z
@@ -74,14 +81,12 @@ class Blob:
         self.delta_D = delta_D
         self.Gamma = Gamma
         self.B = B
-        self._n_e = n_e
-        self._n_p = n_p
+        self._n_e : ParticleDistribution = n_e
+        self._n_p : ParticleDistribution = n_p
         self.xi = xi
 
-        # we might want to have different array of Lorentz factors for e and p
-        self.set_gamma_e(gamma_e_size, self._n_e.gamma_min, self._n_e.gamma_max)
-        if self._n_p is not None:
-            self.set_gamma_p(gamma_p_size, self._n_p.gamma_min, self._n_p.gamma_max)
+        self.gamma_e_size = gamma_e_size
+        self.gamma_p_size = gamma_p_size
 
     @property
     def V_b(self):
@@ -130,24 +135,12 @@ class Blob:
         mu_s = np.cos(theta_s.to("rad").value)
         self.delta_D = 1 / (self.Gamma * (1 - self.Beta * mu_s))
 
-    def set_gamma_e(self, gamma_size, gamma_min=1, gamma_max=1e8):
-        """Set the array of Lorentz factors for the electrons."""
-        self.gamma_e_size = gamma_size
-        self.gamma_e_min = gamma_min
-        self.gamma_e_max = gamma_max
-
-    def set_gamma_p(self, gamma_size, gamma_min=1, gamma_max=1e8):
-        """Set the array of Lorentz factors for the protons."""
-        self.gamma_p_size = gamma_size
-        self.gamma_p_min = gamma_min
-        self.gamma_p_max = gamma_max
-
     @property
     def gamma_e(self):
         """Array of electrons Lorentz factors, to be used for integration in the
         reference frame comoving with the emission region."""
         return np.logspace(
-            np.log10(self.gamma_e_min), np.log10(self.gamma_e_max), self.gamma_e_size
+            np.log10(self._n_e.gamma_min), np.log10(self._n_e.gamma_max), self.gamma_e_size
         )
 
     @property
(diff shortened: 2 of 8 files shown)


agnpy/compton/synchrotron_self_compton.py
```diff
@@ -33,6 +33,36 @@ def __init__(self, blob, ssa=False, integrator=np.trapz):
...
+            self.blob.R_b,
+            self.blob.n_e,
+            *self.blob.n_e.parameters,
+            gamma=gamma
+        )
+
+        # differential energy density
+        u_epsilon0 = (3 * np.power(self.blob.d_L, 2) * sed_flux_synch) / (
+                c * np.power(self.blob.R_b, 2) * np.power(self.blob.delta_D, 4) * epsilon0
+        )
```
jsitarek, line 57: I think you need an extra factor of 3/4 here, compere with this:
https://github.com/cosimoNigro/agnpy/blame/506c038fa751369e63e20a85a222558a8997b8d4/agnpy/compton/synchrotron_self_compton.py#L109-L111

agnpy/emission_regions/blob.py
```diff
@@ -393,3 +397,50 @@ def u_ph_synch(self):
             * np.trapz(np.power(self.gamma_e, 2) * self.n_e(self.gamma_e), self.gamma_e)
         )
         return u_ph.to("erg cm-3")
+
+    def n_e_time_evolution(self, energy_loss_function, time, subintervals_count=10):
+        """Performs the time evolution of the electron distribution inside the blob.
+
+        Parameters
+        ----------
+        energy_loss_function : function
```
jsitarek, line 406: this part you might need to rethink. You specify here a single function, while in reality there will be different processes to be considered and they will have different effect at different moments and at varying energies. So either specify here a list of functions (I guess it is possible??), and loop over them, or a user will need to write a function where he adds a few processes and specifies this function here.

agnpy/emission_regions/blob.py
```diff
@@ -393,3 +397,50 @@ def u_ph_synch(self):
...
+            raise ValueError("subintervals_count must be > 0")
+
+        unit_time_interval = time / subintervals_count
+
+        def gamma_recalculated_after_loss(gamma):
+            old_energy = (gamma * mec2).to("erg")
+            energy_loss_per_time = energy_loss_function(gamma).to("erg s-1")
+            energy_loss = (energy_loss_per_time * unit_time_interval).to("erg")
+            new_energy = old_energy - energy_loss
+            if np.any(new_energy < 0):
```
jsitarek, line 423: this we discussed in person. If the new energy is negative then the code has failed miserably with way too large time step. However if the loss is e.g. 50% is also is very large, because energy losses at E/2 will be different. And limiting the global step size can be inefficient because at low energies nothing is happening. 
The most efficient solution might be try a given step, but at energies where it is too large (e.g. change of energy is more then 10%) the step is divided iteratively by 2 until the change is small enough at this energy.

agnpy/spectra/tests/test_spectra.py
```diff
@@ -760,12 +760,13 @@ def test_compare_numerical_results_with_analytical_calculation(self, p, time_and
         initial_n_e = PowerLaw(k, p, gamma_min=gamma_min, gamma_max=gamma_max, mass=m_e)
         blob = Blob(1e16 * u.cm, Distance(1e27, unit=u.cm).z, 0, 10, 1 * u.G, n_e=initial_n_e)
         synch = Synchrotron(blob)
-        evaluated_n_e = initial_n_e.evaluate_time(time, synch.electron_energy_loss_per_time, subintervals_count=steps)
+        blob.n_e_time_evolution(synch.electron_energy_loss_rate, time, subintervals_count=steps)
+        evaluated_n_e = blob.n_e
```
jsitarek, line 764: be careful with this, because I suspect that this will make a shallow copy of the n_e

agnpy/spectra/tests/test_spectra.py
```diff
@@ -803,11 +804,14 @@ def test_compare_calculation_with_calculation_split_into_two(self):
         synch = Synchrotron(blob)
 
         # iterate over 60 s in 20 steps
-        eval_1 = initial_n_e.evaluate_time(60*u.s, synch.electron_energy_loss_per_time, subintervals_count=20)
+        blob.n_e_time_evolution(synch.electron_energy_loss_rate, 60 * u.s, subintervals_count=20)
+        eval_1 = blob.n_e
```
jsitarek, line 808: another shallow copy in eval_1 and eval_2

agnpy/spectra/tests/test_spectra.py
```diff
@@ -817,4 +821,62 @@ def test_compare_calculation_with_calculation_split_into_two(self):
...
+        )
+
+        blob = Blob(R_b, Distance(1e27, unit=u.cm).z, 10, 10, 1 * u.G, n_e=n_e)
+        synch = Synchrotron(blob)
+        ssc = SynchrotronSelfCompton(blob)
+
+        time = 60 * u.s
+        intervals = 600
+
+        eletronsBefore = n_e.integrate(n_e.gamma_min, n_e.gamma_max)
```
jsitarek, line 842: typo in the variable name: electronsBefore

agnpy/spectra/tests/test_spectra.py
```diff
@@ -817,4 +821,62 @@ def test_compare_calculation_with_calculation_split_into_two(self):
...
+        )
+
+        blob = Blob(R_b, Distance(1e27, unit=u.cm).z, 10, 10, 1 * u.G, n_e=n_e)
+        ssc = SynchrotronSelfCompton(blob)
+
+        time = 60 * u.s
+        intervals = 200
+
+        blob.n_e_time_evolution(ssc.electron_energy_loss_rate_thomson, time, subintervals_count=intervals)
+        n_e_thompson = blob.n_e
```
jsitarek, line 877: possibly a shallow copy here

agnpy/synchrotron/synchrotron.py
```diff
@@ -90,18 +90,13 @@ def __init__(self, blob, ssa=False, integrator=np.trapz):
...
-        fixed = self._electron_energy_loss_formula_prefix()
+    def electron_energy_loss_rate(self, gamma):
+        fixed = self._electron_energy_loss_formula_prefactor()
         return fixed * gamma ** 2
 
-    def _electron_energy_loss_formula_prefix(self):
-        # using SI units here because of the ambiguity of the different CGS systems in terms of mu0 definition
-        magn_energy_dens = (self.blob.B.to("T") ** 2) / (2 * mu0)
-        return ((4 / 3) * sigma_T * c * magn_energy_dens).to("erg/s")
+    def _electron_energy_loss_formula_prefactor(self):
```
jsitarek, line 97: not sure if it makes sense to split it into two functions, especially that you do not do it for IC losses

grzegorzbor: It's split, because the "prefactor"  calculation is done in some test, so I didn't want to duplicate the formulate, and instead extracted it to the (sub-)function for reuse in those tests

agnpy/emission_regions/blob.py
grzegorzbor: Yes I agree, this will need to be changed, but for now it's easier to use it and test it this way, and it's also quite generic. But when we add support for more processes, this part will need some rethinking, how to combine such multiple processes.

agnpy/spectra/tests/test_spectra.py
grzegorzbor: As we already discussed, in this case it's safe because i replace the n_e in blob at each calculation step with a new copy. But just in case the internal implementation changes in the future, i added deep copy here

agnpy/compton/synchrotron_self_compton.py
```diff
@@ -33,6 +33,40 @@ def __init__(self, blob, ssa=False, integrator=np.trapz):
...
+        # factor 3 / 4 for averaging in a sphere - see comment inside the evaluate_sed_flux method for u_synch calculation
+        u_epsilon0 *= 3 / 4
+
+        # formulas from paragraph 2.1, Moderski2005
+        b_ = 4 * gamma.reshape(len(gamma),1) * epsilon0.value.reshape(1,len(epsilon0)) # matrix [gamma x epsilon]
+        b_flattened = b_.reshape(b_.shape[0] * b_.shape[1])
+        f_KN_flattened = np.array([1 / (1 + b) ** 1.5 if b < 1e4 else (9 / (2 * b**2)) * (log(b) - 11/6) for b in b_flattened])
+        f_KN = f_KN_flattened.reshape(len(gamma), len(epsilon0)) # back to [gamma x epsilon] matrix
```
jsitarek, line 65: I think this double reshape and for loop also adds to the execution time of the code. You can do it simpler and probably faster using np.where and operating directly on the matrix

agnpy/spectra/tests/test_spectra.py
```diff
@@ -803,11 +810,14 @@ def test_compare_calculation_with_calculation_split_into_two(self):
...
+        blob.n_e_time_evolution(synch.electron_energy_loss_rate, 60 * u.s, subintervals_count=20)
+        eval_1 = deepcopy(blob.n_e)
         # iterate first over 30 s, and then, starting with interpolated distribution, over the remaining 30 s,
         # with slightly different number of subintervals
-        eval_2 = initial_n_e.evaluate_time(30 * u.s, synch.electron_energy_loss_per_time, subintervals_count=10)\
-            .evaluate_time(30 * u.s, synch.electron_energy_loss_per_time, subintervals_count=8)
+        blob.n_e = initial_n_e
+        blob.n_e_time_evolution(synch.electron_energy_loss_rate, 30 * u.s, subintervals_count=10)
+        blob.n_e_time_evolution(synch.electron_energy_loss_rate, 30 * u.s, subintervals_count=8)
+        eval_2 = deepcopy(blob.n_e)
```
jsitarek, line 820: here in principle you do not need deepcopy, because eval_2 is not being modified anymore

agnpy/spectra/tests/test_spectra.py
```diff
@@ -817,4 +827,63 @@ def test_compare_calculation_with_calculation_split_into_two(self):
...
+        n_e = PowerLaw.from_total_energy(
+            1e48 * u.erg,
+            4 / 3 * np.pi * R_b ** 3,
+            p=2.8,
+            gamma_min=1e2,
+            gamma_max=1e7,
+            mass=m_e,
+        )
+
+        blob = Blob(R_b, Distance(1e27, unit=u.cm).z, 10, 10, 1 * u.G, n_e=n_e)
```
jsitarek, line 841: this blob is being used in a bunch of tests, maybe make it a fixture?

agnpy/spectra/tests/test_spectra.py
```diff
@@ -817,4 +827,63 @@ def test_compare_calculation_with_calculation_split_into_two(self):
...
+
+        time = 60 * u.s
+        intervals = 600
+
+        electronsBefore = n_e.integrate(n_e.gamma_min, n_e.gamma_max)
+
+        blob.n_e_time_evolution(synch.electron_energy_loss_rate, time, subintervals_count=intervals)
+        electronsAfterSync = blob.n_e.integrate(blob.n_e.gamma_min, blob.n_e.gamma_max)
+
+        blob.n_e = n_e
```
jsitarek, line 853: I am wondering if it is enough to only overwrite n_e, or do you need to restore the original gamma factors as well, because those are also modified in the time evolution function

agnpy/spectra/tests/test_spectra.py
```diff
@@ -817,4 +827,63 @@ def test_compare_calculation_with_calculation_split_into_two(self):
...
+        ssc = SynchrotronSelfCompton(blob)
+
+        time = 60 * u.s
+        intervals = 200
+
+        blob.n_e_time_evolution(ssc.electron_energy_loss_rate_thomson, time, subintervals_count=intervals)
+        n_e_thompson = deepcopy(blob.n_e)
+
+        blob.n_e = n_e
+        blob.n_e_time_evolution(ssc.electron_energy_loss_rate, time, subintervals_count=intervals)
```
jsitarek, line 887: this effectively compares that the time_evolution gives the same with two different energy_loss_rate functions that should be close in this energy range. I think it would be better to simply compare the output of electron_energy_loss_rate and ..._thomson on a range of gammas without executing the time_evolution. On the other hand, a test that we were discussing was to compute this analytically (that is relatively simply for Thomson range) and introduce this comparison here.

agnpy/spectra/tests/test_spectra.py
grzegorzbor: fixed, i reset the gamma params in the n_e setter

agnpy/emission_regions/blob.py
```diff
@@ -130,12 +137,24 @@ def set_delta_D(self, Gamma, theta_s):
         mu_s = np.cos(theta_s.to("rad").value)
         self.delta_D = 1 / (self.Gamma * (1 - self.Beta * mu_s))
 
+    def reset_gamma_e(self, gamma_e_size=None):
+        if gamma_e_size is not None:
+            self.gamma_e_size = gamma_e_size
+        self.gamma_e_min = self._n_e.gamma_min
+        self.gamma_e_max = self._n_e.gamma_max
+
     def set_gamma_e(self, gamma_size, gamma_min=1, gamma_max=1e8):
```
jsitarek, line 146: with the reset_gamma_[ep] functions introduced, are the set_gamma_[ep] functions still used anywhere in the code?

agnpy/spectra/tests/test_spectra.py
```diff
@@ -817,4 +824,107 @@ def test_compare_calculation_with_calculation_split_into_two(self):
             eval_2.evaluate(gammas, 1, gamma_min, gamma_max),
             0.001)
 
+    def test_total_number_of_electrons_should_stay_after_time_evolution(self):
+        r_b = 1e16 * u.cm
+        n_e = PowerLaw.from_total_energy(
+            1e48 * u.erg,
+            4 / 3 * np.pi * r_b ** 3,
+            p=2.8,
```
jsitarek, line 832: it might be wise to use a harder spectral index (e.g. -1.5 instead of -2.8) for this test, because this will increase the number of electrons in the high edge of the spectrum that are most affected by the cooling

agnpy/spectra/tests/test_spectra.py
```diff
@@ -817,4 +824,107 @@ def test_compare_calculation_with_calculation_split_into_two(self):
...
+
+        ssc = SynchrotronSelfCompton(blob)
+        blob.n_e_time_evolution(ssc.electron_energy_loss_rate, time, subintervals_count=steps)
+        evaluated_n_e: InterpolatedDistribution = blob.n_e
+
+        def gamma_before(gamma_after_time, before_time):
+            coef = ssc._electron_energy_loss_thomson_formula_prefactor() / (m_e * c ** 2)
+            return 1 / ((1 / gamma_after_time) - before_time * coef)
+
+        def integral_analytical(gamma_bottom, gamma_up):
```
jsitarek, line 911: I think this function is being used in two different test functions - it can be extracted out of the functions to be included only once (and I would change the name to intergral_powerlaw or something like this

grzegorzbor: It turned out there is already sch function in this file, so i removed mine.

agnpy/spectra/tests/test_spectra.py
grzegorzbor: I just tried with 1.5, but  it made the loss rate too high, and the time had to be split into too many subintervals. So i change it to 2.0 and reduced the total time.

agnpy/emission_regions/blob.py
grzegorzbor: Removed. Also, the reset_gamma_[ep] were not really needed, so i removed them too. 
I'd like to remove the gamma_[ep]_size as well, because they don't describe the blob physical parameters. But it needs more work so I'll get back to it later.

agnpy/emission_regions/blob.py
grzegorzbor: I'll work on it in the separate branch.

agnpy/spectra/tests/test_spectra.py
grzegorzbor: Fixed

PR #175, opened 2025-05-24
Heun method implemented

grzegorzbor:


agnpy/emission_regions/blob.py
```diff
@@ -443,14 +458,25 @@ def gamma_recalculated_after_loss(gamma):
...
-            recalc = gamma_recalculated_after_loss(combined)
-            gamma_bins_from = recalc[0::2]
-            gamma_bins_to = recalc[1::2]
-            changed_bins_width = gamma_bins_to - gamma_bins_from
-            if np.any(changed_bins_width <= 0):
-                raise ValueError(
-                    "Energy loss formula returned too big value. Use shorter time ranges.")
+            gamma = interlace(gamma_bins_from, gamma_bins_to)
+            gamma_recalculated = gamma_recalculated_after_loss(gamma)
+            gamma_bins_from_recalc, gamma_bins_to_recalc = deinterlace(gamma_recalculated)
```
jsitarek, line 463: I think lines 463-466 are only needed for Euler method, so they can go into the first if, or even better if you extract it all after if since it is common to the lines in the if for Heun

agnpy/spectra/tests/test_spectra.py
```diff
@@ -937,12 +937,36 @@ def gamma_before_ssc_th(gamma_after_time, before_time):
...
+        euler_reversed_analytically = gamma_before_time(synch1._electron_energy_loss_formula_prefactor, blob1.gamma_e, time)
+
+        blob2 = Blob(n_e=PowerLaw())
+        synch2 = Synchrotron(blob2)
+        blob2.n_e_time_evolution(synch2.electron_energy_loss_rate, time, subintervals_count=steps_heun, method="Heun")
+        heun_reversed_analytically = gamma_before_time(synch2._electron_energy_loss_formula_prefactor, blob2.gamma_e, time)
+
+        errors_heun = np.abs((heun_reversed_analytically - initial_gamma) / initial_gamma)
+        errors_euler = np.abs((euler_reversed_analytically - initial_gamma) / initial_gamma)
+        assert np.alltrue(errors_heun <= errors_euler)
```
jsitarek, line 972: on average it should work like this, but in specific cases you might get by chance a better value with Euler. There are no failed tests in CI, so this one must be passing as well.

agnpy/emission_regions/blob.py
grzegorzbor: as mentioned in the next comment, all these lines are needed to calculate the InterpolatedDistribution which is needed for the "second round" of the Heun method

PR #176, opened 2025-06-02
New optimized method with automatic time splitting

grzegorzbor:


agnpy/radiative_process/tests/test_time_evolution.py
```diff
@@ -0,0 +1,301 @@
...
+        synch = Synchrotron(blob)
+        ssc = SynchrotronSelfCompton(blob)
+
+        time = 6 * u.s
+        intervals = 300
+
+        electrons_before = n_e.integrate()
+
+        TimeEvolution(blob, time, synchrotron_loss(synch)).eval_with_fixed_intervals(
+                                                      intervals_count=intervals, max_change_per_interval=0.5)
```
jsitarek, line 124: the change by 50% is already pretty extreme. This is just a test, but still it would be better to have it in a more realistic setup, maybe 10% or so (also for the lines below)

agnpy/radiative_process/tests/test_time_evolution.py
```diff
@@ -0,0 +1,301 @@
...
+        )
+        k = n_e.k
+        blob = Blob(r_b, n_e=n_e, B=10*u.G)
+        initial_gammas = blob.gamma_e
+
+        ssc = SynchrotronSelfCompton(blob)
+        TimeEvolution(blob, time, ssc_loss(ssc)).eval_with_fixed_intervals(intervals_count=steps)
+        evaluated_n_e: InterpolatedDistribution = blob.n_e
+
+        def gamma_before(prefactor, gamma_after_time, before_time):
```
jsitarek, line 186: this function looks very much like gamma_before_time defined at the beginning

agnpy/radiative_process/tests/test_time_evolution.py
```diff
@@ -0,0 +1,301 @@
...
+        errors_heun = np.abs((heun_reversed_analytically - initial_gamma) / initial_gamma)
+        average_error_euler = np.average(errors_euler)
+        average_error_heun = np.average(errors_heun)
+        print("Average error Euler method", average_error_euler)
+        print("Average error Heun  method", average_error_heun)
+        assert average_error_heun < average_error_euler
+        assert np.alltrue(errors_heun <= errors_euler)
+
+    def test_automatic_and_fixed_intervals_approach_with_single_loop(self):
+        """ Automatic intervals approach with threshold low enough should use just one run of calculations, so should give the same result
```
jsitarek, line 285: shouldn't it be: "with threshold *high* enough" 
the actual threshold is rather low 0.01, but it all depends on the blob parameters

agnpy/radiative_process/time_evolution.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Performs the time evolution of the electron distribution inside the blob, repeating the calculation in the
+        equal-length time intervals.
+
+        Parameters
+        ----------
+        intervals_count : int
+            the calculation will be performed in the N steps of equal duration (time/N),
+            calculating energy change for every gamma value in each step
+            if set to 0 (default), then calculation will use the automatic time splitting, and recalculating energy change rate
```
jsitarek, line 62: something is not right here, the default is 1 not 0, and below you raise an exception if intervals_count<=0

agnpy/radiative_process/time_evolution.py
```diff
@@ -0,0 +1,247 @@
...
+        last_n_array = self._blob.n_e(last_gamma_bins_from)
+        for i in range(intervals_count):
+            last_gamma_bins_from, last_n_array = do_iterative(last_gamma_bins_from, last_n_array, i + 1)
+
+    def eval_with_automatic_intervals(self, method="heun", max_change_per_interval=0.01):
+        """
+        Performs the time evolution of the electron distribution inside the blob, repeating the calculation by the
+        automatically selected time intervals such that the energy change in each interval does not exceed the
+        max_change_per_interval rate. The automatically selected time interval duration may differ per energy bins
+        - shorter intvervals will be used for bins where energy change is faster.
```
jsitarek, line 127: intervals 
(there is a typo)

agnpy/radiative_process/tests/test_time_evolution.py
```diff
@@ -0,0 +1,301 @@
...
+from agnpy.utils.conversion import mec2
+from agnpy.utils.math import power_law_integral
+
+
+def gamma_before_time(prefactor, gamma_after_time, time):
+    """
+    Reverse-time calculation of the gamma value before the energy loss from formula -dE/dt ~ (E ** 2)
+    Applicable to synchrotron and Thomson losses
+    """
+    return 1 / ((1 / gamma_after_time) - time * prefactor() / (m_e * c ** 2))
```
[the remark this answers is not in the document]
grzegorzbor: fixed

agnpy/radiative_process/tests/test_time_evolution.py
grzegorzbor: yes, a leftover from an earlier versions, removed now

agnpy/radiative_process/tests/test_time_evolution.py
grzegorzbor: fixed

agnpy/radiative_process/time_evolution.py
grzegorzbor: yes, also a leftover from the earlier version, removed

agnpy/radiative_process/time_evolution.py
grzegorzbor: fixed

agnpy/utils/math.py
```diff
@@ -117,3 +117,11 @@ def trapz_loglog(y, x, axis=0):
     vals[tozero] = 0.0
 
     return np.add.reduce(vals, axis) * x_unit * y_unit
+
+def power_law_integral(k_e, p, gamma_min, gamma_max):
+    """Analytical integral of the power law."""
```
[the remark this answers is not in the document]
grzegorzbor: done

agnpy/radiative_process/time_evolution.py
```diff
@@ -0,0 +1,247 @@
...
+        The blob for which the time evolution will be performed. As a result of the time evolution,
+        the blob.n_e will be replaced with the InterpolatedDistribution.
+    time : ~astropy.units.Quantity
+         Total time for the calculation
+    energy_change_functions : a function, or an array of functions, of EnergyChangeFnType type
+         The function(s) to be used for calculation of energy change rate per gamma values
+         (for energy gain processes, function should return positive values; for loss, negative values)
+    """
+
+    def __init__(self, blob: Blob, time: Quantity, energy_change_functions: Union[EnergyChangeFnType, Sequence[EnergyChangeFnType]]):
```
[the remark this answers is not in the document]
grzegorzbor: It's only a type hint for code-checking tools, has no impact on the performance, so I think it's fine

agnpy/radiative_process/time_evolution.py
```diff
@@ -0,0 +1,247 @@
...
+        unit_time_interval_sec = (self._time / intervals_count).to("s")
+        # for logging only:
+        fmt_width = len(str(intervals_count))
+
+        def do_iterative(gamma_bins_from, n_array, iteration):
+            """
+            Iterative algorithm for evaluating the electron energy and density. For each gamma point create a narrow bin,
+            calculate the energy change for start and end of the bin, and scale up density by the bin narrowing factor.
+            gamma_bins_from - size N
+            n_array - size N
```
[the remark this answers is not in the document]
grzegorzbor: the n_array? it's the electron density distribution

agnpy/radiative_process/time_evolution.py
```diff
@@ -0,0 +1,247 @@
...
+    @staticmethod
+    def _recalc_gamma_bins_and_density(energy_bins, abs_energy_changes, n_array):
+        energy_bins_from, energy_bins_to = TimeEvolution._deinterlace(energy_bins)
+        energy_bins_width = energy_bins_to - energy_bins_from
+        new_energy_bins = energy_bins + abs_energy_changes
+        new_energy_bins_from, new_energy_bins_to = TimeEvolution._deinterlace(new_energy_bins)
+        new_energy_bins_width = new_energy_bins_to - new_energy_bins_from
+        density_increase = energy_bins_width / new_energy_bins_width
+        new_n_array = n_array * density_increase
+        new_gamma_bins = (new_energy_bins / mec2).value
```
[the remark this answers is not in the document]
grzegorzbor: The energy bins start-end are always expressed in ergs in the algorihtm, so it should be safe. But i changed to .to_value('') anyway, sounds less error-prone in case of changes.

agnpy/radiative_process/time_evolution.py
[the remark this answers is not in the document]
grzegorzbor: Ok, I just modified it to convert the single value to a single-element array right in the constructor, so the performance impact should be close to zero. 
And i think it's still a little bit nicer to use when you can just pass a single function as an argument, without being forced to wrap it explicitly into the array.

agnpy/spectra/spectra.py
```diff
@@ -785,9 +785,10 @@ def evaluate(self, gamma, norm, gamma_min, gamma_max):
             if (gamma[-1] - gamma_max) / gamma_max < 1e-10:
                 gamma[-1] = gamma_max
 
+        mask = (gamma_min <= gamma) * (gamma <= gamma_max)
         values = np.where(
-            (gamma_min <= gamma) * (gamma <= gamma_max),
-            np.power(10, self.log10_interp(log10_gamma)),
+            mask,
+            np.power(10, np.where(mask, self.log10_interp(log10_gamma), 0)),
```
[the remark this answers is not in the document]
grzegorzbor: The additional inner mask was to prevent the overflow warning from the power function. I agree it was tricky so I simplified it now, should be easier to understand

Issue #142, opened 2023-05-03
Problem about load_absorption_table in class EBL


huangylei:
I've found differences in the EBL absorption(with Saldana-Lopez model) between `agnpy` and the data released by Saldana-Lopez. 

Then I checked the function `load_absorption_table` and found that all the models use `np.unique()` to return the sorted unique elements of an array.  However,  we don't need to sort data in Saldana-Lopez model.

cosimoNigro:
@huangylei, thanks a lot for reporting this.
I will make a fix soon.

huangylei:
I used the table that Saldana-Lopez provided, and compared it with the result by `agnpy`. They are different. (Unfortunately 
 I cannot find the code) 

I checked the function `load_absorption_table` into:

```python
def load_absorption_table(self):
        """load the reference values from the table file to be interpolated later"""
        f = fits.open(self.model_file)
        self.energy_ref = (
            np.sqrt(f["ENERGIES"].data["ENERG_LO"] * f["ENERGIES"].data["ENERG_HI"])
            * u.keV
        )
        # Franceschini file has two columns repeated, eliminate them
        if self.model == "franceschini":
            self.z_ref = np.unique(f["SPECTRA"].data["PARAMVAL"])
            self.values_ref = np.unique(f["SPECTRA"].data["INTPSPEC"], axis=0)
        else:
            self.z_ref = f["SPECTRA"].data["PARAMVAL"]
            self.values_ref = f["SPECTRA"].data["INTPSPEC"]
```
It worked and perfectly match the result that Saldana-Lopez provided. After this change, the aborption should be:

<img width="870" alt="截屏2023-05-23 17 10 19" src="https://github.com/cosimoNigro/agnpy/assets/51933895/1ac2cc90-ea13-463c-978d-e57754ba805c">

cosimoNigro:
Fixed by #182.

PR #182, opened 2025-07-24
Fix EBL absorption and interpolation, add a comparison with Gammapy

cosimoNigro:
Hi,

this fix was actually triggered by @HealthyPear's issue on interpolation I corrected.
Then I remembered #142 by @huangylei (thank you!), 
where she complained she found some difference w.r.t the Saldana Lopez model.

When comparing with `Gammapy`, turns out the `agnpy` implementation was not correct, here an example.
image
I don't know if the issue was due to the fact that we were interpolating using the logs of energies,
or to the usage of `np.unique` to eliminate repeated rows in one of the Franceschini files.
Anyhow, I tried to replicate `Gammapy`'s approach, and results look much more compatible
(note `Gammapy` is using a `ScaledRegularGridInterpolator`, built on top of `scipy`'s `RegularGridInterpolator`).

image

I need some more time to fix the failing tests though, some energy or redshift values appear to be considered out of range.


jsitarek:
the differneces that you are showing are mostly in the range that is so heavily absorbed that it does not matter what is the actual value of the attenuation factor. Have you checked how it looks like with low redshifts (like 0.123)?
The values are so small that you should be actually careful with the single vs double precision because you might have some effect from the numerical errors. 

I did not check but I would guess that the different models will probably have somethat different tables (different energy and redshift node values), so then depending if the values at which you do the comparisons is closer or farther from the node values you will be more or less affected by the interpolation issues

huangylei:
Hello you all!

I wonder if you have checked the function I mentioned last time. Maybe it can be helpful? You can review my issue and check the exact meaning of np.unique(). It includes the function of sorting and may affect the table you use.

cosimoNigro:
Ok, the tests passed, if there are no further objections I'd merge.

jsitarek:
Thx @cosimoNigro 
I think the scipy has changed the default type representation of numbers in some operations- we had some issues with this in MCP when upgrading the version. 
Even while the smallest float32 number is 1.e-38, the relative accuracy is something like 1.e-7. In the plot that you show the values are not only very small, but also going down very rapidly. In the tricky part you might have a few orders of magnitude change between one energy and another, so adding and subtracting floats you might run into numerical accuracy problems. The issue would be there if interpolation is done on absorption factor - interpolation on tau makes much more sense and would be more accurate. 
Either way, for all the practical purposes the code seems to work fine, thanks for fixing and merging this.

PR #194, opened 2026-03-19
Started implementation of extended jets

cosimoNigro:
Started to work on implementation of the extended jet model, following [Cotter and Potter 2012](https://ui.adsabs.harvard.edu/abs/2012MNRAS.423..756P/abstract), with @udaisharma99.


agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+class Cone:
+    r"""Uniform Conical Jet Model.
+
+    **Note:** We will assume that the jet has an opening angle θ_{opening}, 
+    initial radius R0 and length L, all defined in the centre of momentum 
+    frame of the fluid.
+
+    Parameters
+    ----------
+    R_o : :class:`~astropy.units.Quantity`
```
cosimoNigro: Since its $R_0$ (R nought) in the paper the subscript should be a zero not an "o".
Replace `R_o -> R_0` in all the occurrences.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        Length of the Jet
+    theta : class:`~astropy.units.Quantity`
+        Opening angle of the cone
+    z : float
+        redshift of the source
+    delta_D : float
+        Doppler factor of the relativistic outflow
+    Gamma : float
+        Lorentz factor of the relativistic outflow
+    B : :class:`~astropy.units.Quantity`
```
cosimoNigro: this parameter should be $B_0$ (i.e. `B_0`)

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        magnetic field at the base of the cone
+    n_e : :class:`~agnpy.spectra.ParticleDistribution`
+        electron distribution contained in the blob
```
cosimoNigro: `in the blob` -> `in the cone`

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+    n_p : :class:`~agnpy.spectra.ParticleDistribution`
+        proton distribution contained in the blob
+    """
+
+## Keeping the default parameters same as defined in Blob (CΟNFIRM FOR NEW PARAMETERS!!!!)
+    def __init__(
+        self,
+        L = 1e18 * u.cm,
+        R_o = 1e14 * u.cm,
```
cosimoNigro: Same as before, replace with `R_0` and `B_0`.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        redshift of the source
+    delta_D : float
+        Doppler factor of the relativistic outflow
+    Gamma : float
+        Lorentz factor of the relativistic outflow
+    B : :class:`~astropy.units.Quantity`
+        magnetic field at the base of the cone
+    n_e : :class:`~agnpy.spectra.ParticleDistribution`
+        electron distribution contained in the blob
+    n_p : :class:`~agnpy.spectra.ParticleDistribution`
```
cosimoNigro: Since we are not computing proton synchrotron for the moment, I suggest removing this parameter from the docstring. Also, you did not initialise any proton distribution later. So, just remove it for the time being.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        xi=1.0,
+        gamma_e_size=200,
+        x_size=100,
+        cosmology=None
+    ):
+        if not isinstance(delta_D, numbers.Number) or delta_D <= 0:
+            raise ValueError("delta_D must be a positive number")
+
+        self.L = L.to("cm")
+        self.R_o = R_o
```
cosimoNigro: A last reminder to replace all the future occurrences of `R_o` and `B_o` with `R_0` and `B_0`.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        self.d_L = Distance(z=self.z, cosmology=cosmology).cgs
+        self.delta_D = delta_D
+        self.Gamma = Gamma
+        self._n_e : ParticleDistribution = n_e          
+        self.xi = xi
+        self.gamma_e_size = gamma_e_size
+        self.x_size = x_size
+    
+    @classmethod
+    def from_jet_power(cls, W_j, **kwargs ):
```
cosimoNigro: Missing docstring. What does this function do?

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        B = kwargs.get("B_o")
+        gamma = kwargs.get("Gamma")
+        u_B = ((B_to_cgs(B)**2)/(8*np.pi)).to('erg cm-3')
+        R_o_squared = (W_j)/(2*u_B*np.pi*(gamma**2)*c.cgs)
+        R_o = np.sqrt(R_o_squared).to('cm')
+        return cls( R_o=R_o , **kwargs)
+        
+## 1. Jet Structure
```
cosimoNigro: Remove this comments between function and function.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        """Radius of the cone as a function of distance from the base"""
+        return self.R_o + self.x * np.tan(self.theta)
+    
+    @property
+    def B_x(self):
+        """Magnetic field of the cone as a function of distance from the base"""
+        return self.B_o * (self.R_o / self.R_x)
+
+    @property
+    def x_cc(self, x_min=0.01 * u.pc):
```
cosimoNigro: just out of curiosity, why `x_cc` what does the CC in underscore stands for?

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        """Spatial grid modified for Chang and Cooper numerical scheme"""
+        return self.x_cc[1:]
+    
+    @property
+    def gamma_e_cc(self):
+        """Array of electrons Lorentz factors, to be used for integration in the
+        reference frame comoving with the emission region."""
+        return np.logspace(
+            np.log10(self._n_e.gamma_min), np.log10(self._n_e.gamma_max), self.gamma_e_size
+        #gamma_max is computed through x_i ; based on Fermi 1st order acceleration coefficient
```
cosimoNigro: I don't understand this comment about`gamma_max`...
we just use `gamma_max` from the electron distribution in the end, no?

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        )
+    
+    @property
+    def gamma_e(self):
+        """Array of electrons Lorentz factors modified for Chang and Cooper numerical method, 
+        to be used for integration in the reference frame comoving with the emission region."""
+        return self.gamma_e_cc[1:-1:2]
+    
+    @property
```
cosimoNigro, line 355: I guess this is something we used for external Compton in the `Blob` class and we don't need it here... right?

electron_evolution.ipynb
cosimoNigro, line 1: Notebook should not go in the `agnpy` dir:
- if used for building and testing a new features they should not be committed or at best go into the `examples` subdir;
- if cleaned, well-documented, and organised they should be properly embedded in the documentation.

numerical_evolution.ipynb
cosimoNigro, line 1: Same comment as for the previous notebook.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+    def gamma_e_external_frame(self):
+        """Array of electrons Lorentz factors, to be used for integration in the
+        reference frame external to the emission region."""
+        return np.logspace(1, 9, self.gamma_e_size)
+
+## 3. Electron Properties: Number Density, Total Number, Energy Density, Total Energy
+    @property
+    def n_e_base(self):
+        """Electron distribution as obtained by agnpy.spectra, with modified normalization 
+        imposed due to equipartition condition (check norm_equi)
```
cosimoNigro: So, every electron distribution is scaled to be at equipartition?
Shouldn't this be optional, or is this a prescription of Potter's and Cotter's model?
(i.e. $U_e = U_B$ always).

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        units: cm-3
+        """ 
+        return (self._n_e(self.gamma_e_cc)[1:-1:2])*self.norm_equi*u.cm**-3 *np.pi * self.R_o**2
+
+    @property
+    def norm_equi(self):
```
cosimoNigro: Should go before other functions calling it.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        """Modified normalization to convolve with input electron distribution
+        to impose equipartition. Obtained by equating magnetic energy density at the base to 
+        electron energy density. 
+        units: cm-3"""
+        e_energy_initial = mec2.to('eV') * np.trapz(self.gamma_e_cc * self._n_e(self.gamma_e_cc), self.gamma_e_cc)
```
cosimoNigro: Any particle distribution has already an `integral` method, with a `gamma_power` argument that allow you to multiply the integrand by energy, see https://github.com/cosimoNigro/agnpy/blob/master/agnpy/spectra/spectra.py#L54.

agnpy/emission_regions/cone.py
cosimoNigro: Got it on a second read, it's Chang Cooper!

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+    @property
+    def W_e(self):
+        r"""Total energy of electrons"""
+        W_e = mec2.cgs * np.trapz(self.gamma_e * self.N_e_gamma, self.gamma_e)
+        return W_e
+    
+    @property
+    def N_e(self): 
+        r"""Total Number of electrons in the Jet,
+        :math:`N_{\rm p}(\gamma') = V_{\rm b}\,n_{\rm p}(\gamma')`.
```
cosimoNigro: This formula still refers to the blob.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+        mu_s = np.cos(theta_s.to("rad").value)
+        self.delta_D = 1 / (self.Gamma * (1 - self.Beta * mu_s))
+
+    def __str__(self):
+        """Printable summary of the Conical Jet."""
+        resume = (
+            "* Conical emission region\n"
+            + f" - R_o (Radius at the base of the cone): {self.R_o.cgs:.2e}\n"
+            + f" - L (Length of the Jet): {self.L.cgs:.2e}\n"
+            + f" - θ (Opening Angle of the Conical Jet): {self.theta:.2e}\n"
```
cosimoNigro: I'm not sure this $\theta$ will be correctly displayed in the terminal.
See https://docs.python.org/3/howto/unicode.html.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+            + f" - d_L (source luminosity distance):{self.d_L.cgs:.2e}\n"
+            + f" - delta_D (blob Doppler factor): {self.delta_D:.2e}\n"
+            + f" - Gamma (Bulk Lorentz factor): {self.Gamma:.2e}\n"
+            + f" - Beta (Bulk relativistic velocity): {self.Beta:.2e}\n"
+            + f" - theta_s (jet viewing angle): {self.theta_s:.2e}\n"
+            + f" - B (Magnetic field at the base the jet): {self.B_o:.2e}\n"
+            + f" - xi (coefficient for 1st order Fermi acceleration) : {self.xi:.2e}\n"
+        )
+        
+
```
cosimoNigro: A single newline should be left at the end of the file.

agnpy/synchrotron/tests/test_synchrotron_cone.py
```diff
@@ -0,0 +1,108 @@
+# test the synchrotron module
```
cosimoNigro: `# test the synchrotron_cone module`

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+from matplotlib import cycler, rcParams
+from agnpy.utils.conversion import nu_to_epsilon_prime, B_to_cgs, lambda_c_e, mec2
+import numpy as np
+import astropy.units as u
+from astropy.coordinates import Distance
+from scipy.integrate import cumulative_trapezoid
+from astropy.constants import c, sigma_T, m_e
+
+from .. import InterpolatedDistribution, ParticleDistribution
+from agnpy.spectra.spectra import ExpCutoffPowerLaw
```
cosimoNigro: Do we use these distributions at all?
I see only `PowerLaw` being used.

agnpy/emission_regions/cone.py
```diff
@@ -0,0 +1,278 @@
...
+
+import matplotlib.pyplot as plt
+from astropy.constants import c, sigma_T, m_e, e, mu0
+from scipy.sparse import diags
+from matplotlib import cycler, rcParams
+from agnpy.utils.conversion import nu_to_epsilon_prime, B_to_cgs, lambda_c_e, mec2
+import numpy as np
+import astropy.units as u
+from astropy.coordinates import Distance
+from scipy.integrate import cumulative_trapezoid
```
cosimoNigro: `cumulative_trapezoid` seem not to be used in this submodule

agnpy/synchrotron/tests/test_synchrotron_cone.py
```diff
@@ -0,0 +1,108 @@
...
+
+agnpy_dir = Path(__file__).parent.parent.parent  # go to the agnpy root
+# where to read sampled files
+data_dir = agnpy_dir / "data"
+# where to save figures, clean-up before making the new
+figures_dir = clean_and_make_dir(agnpy_dir, "crosschecks/figures/synchrotron")
+
+# Parameters used in Table 1 of Potter and Cotter; Fitting BL Lac (abdo et al. 2011)
+z=0.0686
+B_o =  (3.63 * 1e-5 * u.T).to(u.G)
```
cosimoNigro: Same as before: use `B_0` and `R_0`.

agnpy/synchrotron/tests/test_synchrotron_cone.py
```diff
@@ -0,0 +1,108 @@
...
+u_B = ((B_to_cgs(B_o)**2)/(8*np.pi)).to('erg cm-3')
+R_o = 7.32 * 1e15 * u.cm
+W_j = 7.63 * 1e36* u.W
+L = (100 * u.pc)
```
cosimoNigro: No need for the parentheses.

agnpy/synchrotron/tests/test_synchrotron_cone.py
```diff
@@ -0,0 +1,108 @@
...
+            "agnpy",
+            "CnP-Paper I: Fig 3b",
+            figure_title,
+            figure_path,
+            "sed",
+            comparison_range=nu_range.to_value("Hz"),
+        )
+
+        # requires that the SED points deviate less than 5% from the figure
+        assert check_deviation(nu_ref, sed_agnpy, sed_ref, 0.25, nu_range)
```
cosimoNigro: leave a new empty line at the end of each file.

agnpy/synchrotron/tests/test_synchrotron_cone.py
```diff
@@ -0,0 +1,108 @@
...
+A_equi = 1.0
+# Electron Distribution imposed by Potter and Cotter
+p = 2
+gamma_minimum = (5.11 * 1e6 * u.eV / mec2.to("eV")).value
+gamma_c = (5.6 * 1e9 * u.eV / mec2.to("eV")).value
+n_e = ExpCutoffPowerLaw(k = 1 , mass=m_e, p=p, gamma_min = 0.1*gamma_minimum, gamma_max= 100*gamma_c, gamma_c= gamma_c )
+
+
+class TestConeSynchrotron:
+    @pytest.mark.parametrize(
```
cosimoNigro, line 53: I don't understand the use of the `parametrize`, are we producing one more than one figure?
`parametrize` is used when we want to run the same test with varying parameters
https://docs.pytest.org/en/7.1.x/example/parametrize.html

agnpy/synchrotron/tests/test_synchrotron_cone.py
```diff
@@ -0,0 +1,108 @@
...
+        ],
+    )
+    def test_ssa_reference_sed(
+        self,
+        file_ref,
+        n_e,
+        figure_title,
+        figure_path,
+    ):
+        """Test SSA SED generated by a given electron distribution against the
```
cosimoNigro, line 71: docstring is incorrect.

agnpy/synchrotron/tests/test_synchrotron_cone.py
```diff
@@ -0,0 +1,108 @@
...
+        sed_agnpy = synch.sed_flux(nu_ref, ssa=True)
+
+        # sed comparison plot, we will check between 10^(11) and 10^(19) Hz
+        nu_range = [1e8, 1e19] * u.Hz
+        make_comparison_plot(
+            nu_ref,
+            sed_agnpy,
+            sed_ref,
+            "agnpy",
+            "CnP-Paper I: Fig 3b",
```
cosimoNigro, line 104: I would use `Cotter and Potter I: Fig. 3b`.

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
-# module containing the synchrotron radiative process
 import numpy as np
+from agnpy.emission_regions import Cone, Blob
```
cosimoNigro: Why do we import the `Cone` in the `synchrotron_blob` class?

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
+from ..utils.conversion import nu_to_epsilon_prime, B_to_cgs, lambda_c_e, mec2
 from ..radiative_process import RadiativeProcess
-
-
-__all__ = ["R", "nu_synch_peak", "Synchrotron"]
-
 e = e.gauss
-B_cr = 4.414e13 * u.G  # critical magnetic field
 
+# ============================================================
```
cosimoNigro: Remove this comments between function and function please.

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
-    """Eq. 7.45 in [Dermer2009]_, angle-averaged integrand of the radiated power, the
-    approximation of this function, given in Eq. D7 of [Aharonian2010]_, is used.
-    """
-    term_1_num = 1.808 * np.power(x, 1 / 3)
-    term_1_denom = np.sqrt(1 + 3.4 * np.power(x, 2 / 3))
-    term_2_num = 1 + 2.21 * np.power(x, 2 / 3) + 0.347 * np.power(x, 4 / 3)
-    term_2_denom = 1 + 1.353 * np.power(x, 2 / 3) + 0.217 * np.power(x, 4 / 3)
-    return term_1_num / term_1_denom * term_2_num / term_2_denom * np.exp(-x)
-
+def nu_obs_to_nu_fluid(nu_obs, z, delta_D):
```
cosimoNigro: This function is useful, perhaps it could go here
https://github.com/cosimoNigro/agnpy/blob/master/agnpy/utils/conversion.py
in the `conversion` submodule.

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
 def nu_synch_peak(B, gamma, mass=m_e):
-    """observed peak frequency for monoenergetic electrons
-    Eq. 7.19 in [DermerMenon2009]_"""
+    """Critical synchrotron frequency (Dermer 2009 Eq. 7.19)."""
     B = B_to_cgs(B)
-    nu_peak = (e * B / (2 * np.pi * mass * c)) * np.power(gamma, 2)
+    nu_peak = (3 * e * B / (4 * np.pi * mass * c)) * np.power(gamma, 2)
     return nu_peak.to("Hz")
 
+def Z(eta):
```
cosimoNigro: Why was `R(x)` renamed to `Z(eta)`?
Any specific reason?

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
+    nu_fluid = nu_obs * (1+ z) / delta_D
+    return nu_fluid 
 
 def nu_synch_peak(B, gamma, mass=m_e):
-    """observed peak frequency for monoenergetic electrons
-    Eq. 7.19 in [DermerMenon2009]_"""
+    """Critical synchrotron frequency (Dermer 2009 Eq. 7.19)."""
     B = B_to_cgs(B)
-    nu_peak = (e * B / (2 * np.pi * mass * c)) * np.power(gamma, 2)
+    nu_peak = (3 * e * B / (4 * np.pi * mass * c)) * np.power(gamma, 2)
```
cosimoNigro: Why the new `4/3` factor?

Image

This is the Eq. 7.19 I'm referring to.

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
-    return x.to_value("")
-
-
-def epsilon_B(B):
-    r""":math:`\epsilon_B`, Eq. 7.21 [DermerMenon2009]_"""
-    return (B / B_cr).to_value("")
+    return eta.value
+
+def eta(nu_fluid, nu_peak):
+    """Calculates ration of frequency to critical synchrotron frequency
```
cosimoNigro: `ration`-> `ratio`

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
+    for Conical Jet Model"""
+    eta = nu_fluid / nu_peak
+    return eta
+
+def P_sync(
+        nu_fluid,         # Ratio b/w Observed Frequency Array and Critical Frequency
```
cosimoNigro: Don't need to specify what the arguments are meant for with a comment, when you have the docstring right below.

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
+    B_x : array-like
+        Magnetic field along the jet, shape (N_x,1,1)
+    x : array-like
+        Spatial coordinate along the jet (cm), shape (N_x,)
+    gamma_e : array-like
+        Electron Lorentz factors, shape (1,N_gamma,1)
+    N_e_xg : array-like
+        Electron distribution, shape (N_x,N_gamma,1)
+
+    Returns
```
cosimoNigro: Why returning 3 different objects?
Are they all needed for the calculations?

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
+    gamma_integral = integrator(
+            B_x * N_e_xg * Z_eta,
+            gamma_e,
+            axis=1
+            )                                        
+    # --- prefactor √3 e³ / (m c²) ---
+    prefactor = (np.sqrt(3) * (e)**3 / (mec2.cgs)).cgs
+    emission = prefactor * gamma_integral  
+    P_synch = integrator(emission, x, axis=0)
+    return nu_peak, emission, Z_eta, P_synch.to("erg Hz-1 s-1")
```
cosimoNigro: You actually return 4 objects.
Was this done for debugging?

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -1,74 +1,165 @@
...
+
+
+# ============================================================
```
cosimoNigro: Again, remove this, please.

agnpy/synchrotron/synchrotron_blob.py
```diff
@@ -274,4 +365,4 @@ def sed_peak_nu(self, nu):
         SED peaks
         """
         idx_max = self.sed_flux(nu).argmax()
-        return nu[idx_max]
+        return nu[idx_max]
```
cosimoNigro: New line at the end of file missing.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+from ..utils.math import axes_reshaper, gamma_e_to_integrate
+from ..utils.conversion import nu_to_epsilon_prime, B_to_cgs, lambda_c_e, mec2
+from ..radiative_process import RadiativeProcess
+e = e.gauss
+
+# ============================================================
+# Utility Functions
+# ============================================================
+
+def nu_obs_to_nu_fluid(nu_obs, z, delta_D):
```
cosimoNigro: Duplicate function, you defined it in `synchrotron_blob` already.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
+import numpy as np
+from agnpy.emission_regions import Cone, Blob
+import astropy.units as u
+from astropy.constants import e, h, c, m_e, sigma_T, mu0
+from ..utils.math import axes_reshaper, gamma_e_to_integrate
+from ..utils.conversion import nu_to_epsilon_prime, B_to_cgs, lambda_c_e, mec2
+from ..radiative_process import RadiativeProcess
+e = e.gauss
+
+# ============================================================
```
cosimoNigro: Remove, thanks.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+# Utility Functions
+# ============================================================
+
+def nu_obs_to_nu_fluid(nu_obs, z, delta_D):
+    """Convert observed frequency to comoving (fluid) frame."""
+    nu_fluid = nu_obs * (1+ z) / delta_D
+    return nu_fluid 
+
+def nu_synch_peak(B, gamma, mass=m_e):
```
cosimoNigro, line 14: Same as above.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+    """Critical synchrotron frequency (Dermer 2009 Eq. 7.19)."""
+    B = B_to_cgs(B)
+    nu_peak = (3 * e * B / (4 * np.pi * mass * c)) * np.power(gamma, 2)
+    return nu_peak.to("Hz")
+
+def Z(eta):
```
cosimoNigro: Same.
Perhaps, if the **same** functions are needed by both `synchrotron_blob` and `synchrotron_cone`, they might go in `synchrotron_base` and be imported from there.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+        shape (N_x, N_nu)
+    kernel_Z : ndarray
+        Synchrotron kernel Z(η), shape (N_x, N_gamma, N_nu)
+    L_nu_fluid : Quantity
+        Total luminosity in fluid frame (integrated over x), [erg s⁻¹ Hz⁻¹]
+
+    Notes
+    -----
+    Implements Eq. 7.44 from Dermer (2009).
+    P_mu_fluid = ∫ dx  [ √3 e^3 B(x) / (m_e c^2) ] ∫ d(gamma) N_e(gamma,x) Z(eta = mu/mu_c)
```
cosimoNigro: Use proper math formatting for the docstring, see
https://ikerdocs-sphinx.readthedocs.io/syntax/math.html.

By using this format, equations will be nicely rendered by sphynx when the docs are build, e.g.

Image

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+    #Calculate Eta
+    eta_ = eta(nu_fluid,nu_peak)
+    Z_eta = Z(eta_)
+    # --- integrate over gamma ---
+    gamma_integral = integrator(
+            B_x * N_e_xg * Z_eta,
+            gamma_e,
+            axis=1
+            )                                        
+    # --- prefactor √3 e³ / (m c²) ---
```
cosimoNigro: Same comment as before, let's avoid special characters in the comments.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+    def Flux_xv(self):
+        return self.F_xv_obs
+    
+    def Bessel(self):
+        return self.Z_eta
+    
+    def nu_peak(self):
+        return self.nu_p
+
+# ============================================================
```
cosimoNigro: Remove, thx!

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+    emission = prefactor * gamma_integral  
+    P_synch = integrator(emission, x, axis=0)
+    return nu_peak, emission, Z_eta, P_synch.to("erg Hz-1 s-1")
+
+
+class SynchrotronCone(RadiativeProcess):
+    """
+    Synchrotron emission from a conical jet.
+
+    Produces:
```
cosimoNigro: Not a proper docstring.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+    - F_nu(x)    → spatially resolved flux
+    - nuF_nu      → SED
+	"""
+
+    def __init__(self, emitter, ssa=False, integrator=np.trapz):
+        self.emitter = emitter
+        self.ssa = ssa
+        self.integrator = integrator
+    
+    def flux_obs(self, nu_obs):
```
cosimoNigro: try to be consistent with the terminology adopted for other radiative processes, e.g. `sed_flux` should be the function that produces the SED in the $\nu F_{\nu}$ format. See
https://github.com/cosimoNigro/agnpy/blob/master/agnpy/synchrotron/synchrotron.py#L229

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+        nu_p, L_xv_fluid, self.Z_eta, L_v_fluid = P_sync(nu_fluid_reshaped, 
+                     B_x_reshaped, 
+                     self.emitter.x, 
+                     gamma_e_reshaped, 
+                     N_e_xg_reshaped,
+                     integrator=np.trapz) #.value * (u.erg/(u.s * u.Hz))
+        self.nu_p = nu_p
+        L_v_obs = L_v_fluid * self.emitter.delta_D**3
+        L_xv_obs = L_xv_fluid * self.emitter.delta_D**3
+        self.F_v_obs = L_v_obs / (4 * np.pi * np.power(self.emitter.d_L, 2))
```
cosimoNigro: Why are they assigned to this class attribute instead of being returned?

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+        self.F_xv_obs = L_xv_obs / (4 * np.pi * np.power(self.emitter.d_L, 2))
+        return self.F_v_obs
+    
+    def Flux_xv(self):
```
cosimoNigro: I don't understand this function...
You get the same output if you do

`cone.Flux_xv`
or 
`cone.F_xv_obs`

why the duplication?

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+        return self.F_xv_obs
+    
+    def Bessel(self):
```
cosimoNigro: Same here, was this some leftover of debugging?

agnpy/synchrotron/synchrotron_cone.py
cosimoNigro: Also if this is returning a $F_{\nu}$ as a function of x, then better to use the variable
`F_nu_x` or directly `emissivity`, as done in the `synchrotron` class
https://github.com/cosimoNigro/agnpy/blob/master/agnpy/synchrotron/synchrotron.py#L191.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+
+        Returns
+        -------
+        kappa_nu : ndarray
+            SSA opacity along the jet, shape (Nx, N_nu)
+
+            ::math::
+        kappa_nu(x) = - frac{1}{8 \pi m_e nu^2} \int d\gamma \, \gamma^2 \, frac{\partial}{\partial \gamma} \left[ frac{N_e(\gamma, x) \, P(nu, \gamma, x)}{\gamma^2} \right]
+        """
+        n_e_xg_reshaped = self.emitter.N_e_xg[:, :, None]/(np.pi * (self.emitter.R_x[:,None,None])**2)
```
cosimoNigro: There is an `axes_reshaper` function to do this.
https://github.com/cosimoNigro/agnpy/blob/master/agnpy/utils/math.py#L22

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+        self.attenuation = tau_to_attenuation(tau_nu)      # reduces flux
+        return self.attenuation  # shape (N_nu)
+
+    def sed_flux(self, nu_obs, ssa=False):
+        self.sed = self.F_v_obs * nu_obs
+        if ssa:
+            self.sed *= self.attenuation
+        return self.sed.to('erg cm-2 s-1')
+    
+    def sed_spatial(self, nu_obs, ssa=False):
```
cosimoNigro: `sed_flux_x`?

So we understand it's an SED flux but along x.

agnpy/synchrotron/synchrotron_cone.py
```diff
@@ -0,0 +1,242 @@
...
+        self.sed_x = self.F_xv_obs * nu_obs
+        if ssa:
+            self.sed_x *= tau_to_attenuation(self.kappa_xv.cgs.value)
+        return self.sed_x.to('erg cm-3 s-1')
+    
+    def sed_peak_flux(self):
+        """#provided a grid of frequencies nu, returns the peak flux of the SED
```
cosimoNigro: Remove `#`from the docstring, same below.

agnpy/synchrotron/tests/test_synchrotron_cone.py
cosimoNigro: Incidentally, it's also `Potter and Cotter`
https://ui.adsabs.harvard.edu/abs/2012MNRAS.423..756P/abstract

agnpy/utils/numerical_methods.py
```diff
@@ -0,0 +1,87 @@
...
+            escape_term = coeff * (dt / t_esc).decompose().value
+            loss_term = (dt * gamma_loss(self.gamma_midpts,B_t)[1:]).decompose().value / self.delta_gamma
+            V2 = 1 + loss_term + escape_term
+            V3 = -(dt * gamma_loss(self.gamma_midpts,B_t)[1:]).decompose().value / self.delta_gamma
+            cc_matrix = diags([V2, V3], offsets=[0, 1]).toarray()
+            N_next = np.linalg.solve(cc_matrix, N_prev)
+            solutions[f"{elapsed_time}"] = N_next 
+            N_prev = N_next
+        N_e_tg_list = []
+        for key in solutions:
```
cosimoNigro, line 88: Why do we need this secondary loop and solutions can't be directly appended within the main loop (I guess after the assignment `N_prev = N_next`)?

cosimoNigro (commented):
Thanks @udaisharma99 for starting this big contribution!

I left a first review. I mostly focused on the code syntax and organisation.
I'll review the detailed calculations in a second moment.

Here are some general comments and next tasks:

### syntax & formatting
- please use [black](https://github.com/psf/black) to automatically format the code;
- remove all the comments in between functions / classes, like
```
   ##########
   # here we do this
   ##########
```
  they feel ChatGPT-ish 😜;
- choose and then stick to a single variable naming convention: quantities with underscore $\nu$ sometimes are reported as `_nu`, sometimes as `_v`. Try to be as consistent as possible with the naming conventions of other classes (e.g. `sed_flux` is the function that returns $\nu F_{\nu}$). Same for the quantity that depends on the distance. We might agree to add a simple `_x` to the names used in other modules (hence `sed_flux_x`, for example for the SED flux along the jet).
 
### structure & organisation
- check the imports, many times you imported functions you did not actually use. `vscode` should automatically detect this and functions and variables not used have a darker colour
Image

- remove duplicate functions. `synchrotron_cone.py` and `synchrotron_blob.py` implement the very same functions. If a function is needed by both classes it can go in another submodule (e.g. `synchrotron_base.py` from which both import.


### tests
- thanks for adding the tests for the synchrotron radiation. Their idea is to cover as much as possible what is implemented. The `Cone` emission region is untested now. You can check from the simplest functionalities (e.g. check if you compute correctly the volume of a truncated cone) to the more complex one (e.g. check against the electron spectra of Potter and Cotter).

cosimoNigro:
A comment that I left out from the review is the `ChangCooperSolver`, if it's a generic class for time evolution, then perhaps it should be incorporated in the modules that @grzegorzbor introduced.

agnpy/emission_regions/cone.py
udaisharma99: Fixed in all cases!

agnpy/synchrotron/synchrotron_cone.py
udaisharma99: Yes, that's right. I removed them now.

agnpy/synchrotron/synchrotron_cone.py
udaisharma99: Done!

agnpy/utils/numerical_methods.py
udaisharma99: It was just to convert the dictionary into a an array but optimized it now to store them directly.

agnpy/synchrotron/synchrotron_cone.py
udaisharma99: Yes, I used it now where I could but the problem is, a lot of reshaping is for 2D arrays like N_e_xg where the axes_reshaper doesn't work. Do you want to use axes_reshaper wherever possible and the broadcasting used above at other places?

agnpy/emission_regions/cone.py
udaisharma99: I modified it to allow the user to set the equipartition parameter. If not specified, it defaults to 1.

agnpy/emission_regions/cone.py
udaisharma99: I fixed this to keep a single x array now, the initial electron distribution is included in the Chang and Cooper output electron distribution array N_e_xg

agnpy/emission_regions/cone.py
udaisharma99: Yes, i am not sure why. Removed it now!

agnpy/emission_regions/cone.py
udaisharma99: I think so, removed for now!

agnpy/emission_regions/cone.py
udaisharma99: To use the same length of gamma array as used everywhere in the calculations, I do not use the integral method. The integral method uses 200 as a fixed gamma_size?

agnpy/synchrotron/synchrotron_blob.py
udaisharma99: To avoid confusion with R(x) used in cone emission region. Fixed for both synchrotron, they now use the same Z(eta) function.

agnpy/synchrotron/synchrotron_blob.py
udaisharma99: Yes, was for debugging. It's fixed now.

agnpy/synchrotron/synchrotron_blob.py
udaisharma99: I calculate Z once so that it can be reused, since it's computationally expensive to calculate. The other 2 are to get the total SED and SED as a function of x.

agnpy/synchrotron/synchrotron_cone.py
udaisharma99: Ah okay I didn't know they are subsequently used for this. Will do, thank you!

agnpy/synchrotron/synchrotron_blob.py
udaisharma99: Done

agnpy/synchrotron/synchrotron_blob.py
udaisharma99: image

I used equation 7.38.

From what I understand, equation 7.38 is a more accurate representation of critical frequency whereas, equation 7.19 is obtained as a qualitative estimate. Maybe we can discuss it once.

agnpy/synchrotron/synchrotron_cone.py
udaisharma99: I have assigned few variables as self attributes so that they are not computed again and again. For example, the synchrotron kernel is very computationally expensive so once it is calculated we store it as a self attribute and can be used again later, wherever required, for instance in calculation of SSA.

agnpy/synchrotron/synchrotron_cone.py
udaisharma99: Emissivity would be the emission per unit volume which is calculated in Blob. but for Cone, we use the electrons per cm as the electron distribution, so emissivity might not be accurate. However, I have cleaned up the code and made a few more changes to make the terminology a bit more transparent and easy to understand.

PR #198, opened 2026-06-04
Ver 0.5.1

grzegorzbor:
I prepared version 0.5.1. 

It's purely the bug-fix relase, for 3 bugs related to project dependencies:

- in version 0.5.0 there was a faulty "import pygments.styles" added by mistake in the timeevolution python file, making the timeevolution code failing if you didn't have this dependency installed (and this dependency is not really used by agnpy)
- version 0.5.0 required both gammapy and sherpa to be installed if you want to use any of the fitting wrappers, which is not correct - you should not need sherpa if you only use gammapy wrapper, and vice-versa.
-  class DistributionToSinglePointCollapseError should be public, so should be in agnpy.time_evolution.types instead of agnpy.time_evolution._time_evolution_utils

These 3 problems are fixed in 0.5.1

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -79,5 +79,5 @@
     ],
     "title": "agnpy",
     "upload_type": "software",
-    "version": "0.5.0"
+    "version": "0.5.1"
 }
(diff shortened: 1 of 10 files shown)


cosimoNigro:
Looks good, thanks @grzegorzbor!

PR #199, opened 2026-07-11
A small fix to prevent interpolation crashes in some very rare cases + project dependencies fix + fixed swapped params

grzegorzbor:
1. A small fix to a special situation when calculating average of equal gamma points returns a different value, which may cause interpolation crashing in some corner-cases.
2. Fix in the project dependencies (added constraints to prevent failures from the unsupported latest versions of dependencies)
3. Fixed swapped parameters

### DIFF
--- a/agnpy/time_evolution/_time_evolution_utils.py
+++ b/agnpy/time_evolution/_time_evolution_utils.py
@@ -196,10 +196,16 @@ def merge_points(x, y, merge_groups, interpolated_distribution):
     - x_new, y_new: arrays with merged points
     """
 
+    def merge_func(x_points):
+        if x_points[0] == x_points[-1]: # if all points are the same, just return the first one (they are sorted so just compare first to last)
+            return x_points[0]
+        else: # else calculate the average
+            return np.power(10, np.mean(np.log10(x_points)))
+
     labels = np.full(len(x), -1)
     for g_idx, group in enumerate(merge_groups):
         labels[group] = g_idx
-    x_means = np.array([np.exp(np.mean(np.log(x[labels == g]))) for g in range(len(merge_groups))])
+    x_means = np.array([merge_func(x[labels == g]) for g in range(len(merge_groups))])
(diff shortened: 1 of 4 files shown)


agnpy/time_evolution/_time_evolution_utils.py
```diff
@@ -196,10 +196,16 @@ def merge_points(x, y, merge_groups, interpolated_distribution):
     - x_new, y_new: arrays with merged points
     """
 
+    def merge_func(x_points):
+        if x_points[0] == x_points[-1]: # if all points are the same, just return the first one
```
jsitarek: this checks first and last point, but you might have merging of multiple points at the same time, and they might not end up as sequential. THe code is probably still fine, because even in this case you would expect possible other points to also be very closeby, but the comment is a bit misleading

agnpy/time_evolution/time_evolution.py
```diff
@@ -270,8 +270,8 @@ def evaluate(self) -> TimeEvaluationResult:
             total_time_elapsed += step_time
             if self._distribution_change_callback is not None:
                 self._distribution_change_callback(TimeEvaluationResult(total_time_elapsed,
-                    gm_bins[0], density, subgroups_density, energy_changes_lb(en_chg_rates), abs_injection_rates,
```
jsitarek, line 273: this is another bugfix that is not mentioned in the PR text

agnpy/time_evolution/_time_evolution_utils.py
grzegorzbor: They are sorted, so if first == last, then all are equal. Added a comment in the code

agnpy/time_evolution/time_evolution.py
grzegorzbor: corrected

jsitarek (approved):
looks good, as soon as the automatic tests pass we will merge it.

PR #200, opened 2026-07-18
Blob expansion implemented

grzegorzbor:
It covers 3 effects: 
1. electron density dilution
2. adiabatic electron energy loss
3. magnetic field decay


docs/time_evolution.rst
```diff
@@ -145,10 +145,48 @@ indicate particle loss).
...
+where the expansion speed :math:`v_{exp}` must be non-negative and lower than the speed of light. Three physical
+consequences of the expansion are then modelled consistently:
+
+* **adiabatic energy losses**: every electron loses energy at the rate :math:`dE/dt = -E \, v_{exp} / R`.
+  This is registered internally as an additional energy change function under the reserved key ``"adiabatic_expansion"``;
+* **density dilution**: the particle density decreases as the volume grows, :math:`dn/dt = -3 \, n \, v_{exp} / R`,
+  which conserves the total number of particles :math:`N = n V_b`. This is registered internally as an additional
+  relative injection (escape) function under the reserved key ``"expansion_dilution"``;
+* **magnetic field decay**: the field follows :math:`B(t) = B_0 (R_0 / R(t))^m`, where the index `m` is set by
+  the `magnetic_field_index` parameter of `BlobExpansion`. Use `m=1` for conservation of the toroidal field flux
```
jsitarek: I am not sure if this part about toroidal/poloidal is correct. Either way, m=1 corresponds to the case when the total energy of magnetic field through the cross section of the blob is constant

docs/time_evolution.rst
```diff
@@ -145,10 +145,48 @@ indicate particle loss).
...
+Because the internal functions are registered alongside the user-provided ones, the automatic step-duration
+constraints (`max_energy_change_per_interval` etc.) also bound the radius growth per step, and the expansion rates
+are reported in the evaluation results under their reserved keys (which must not be used as names of user functions).
+
+As a side effect (in addition to replacing `blob.n_e`), the `evaluate` method advances `blob.R_b` and `blob.B`
+along with the elapsed time. Since other processes, such as synchrotron or SSC losses, read the blob properties
+at every step, they automatically respond to the growing radius and the decaying magnetic field. A subsequent
+`evaluate` call continues the expansion from the radius already reached, so a simulation can be split into
+multiple runs. The `expansion` parameter is not supported together with `optimize_recalculating_slow_rates`,
+because with an expanding blob all rates depend on the radius and the magnetic field, which change at every step.
```
jsitarek: but it might still be that e.g. highest energies force a very small step in which the size of the blob changes little

docs/time_evolution.rst
```diff
@@ -145,10 +145,48 @@ indicate particle loss).
...
+
+.. plot:: snippets/time_evolution_2.py
```
jsitarek: maybe call it time_evolution_expanding.py or something like this

agnpy/time_evolution/tests/test_blob_expansion.py
```diff
@@ -0,0 +1,210 @@
...
+                      step_duration=3 * u.s).evaluate()
+
+        blob2 = Blob(r_0, B=1 * u.G, n_e=deepcopy(initial_n_e))
+        synch2 = synchrotron_loss(Synchrotron(blob2))
+        TimeEvolution(blob2, 30 * u.s, synch2, expansion=expansion, step_duration=3 * u.s).evaluate()
+        TimeEvolution(blob2, 30 * u.s, synch2, expansion=expansion, step_duration=3.5 * u.s).evaluate()
+
+        assert u.isclose(blob1.R_b, blob2.R_b, rtol=1e-9)
+        assert u.isclose(blob1.B, blob2.B, rtol=1e-9)
+        gamma_min, gamma_max = blob1.n_e.gamma_min, blob1.n_e.gamma_max
```
jsitarek: you might get here some errors at the edge if gamma_min is slightly smaller (or gamma_max is slightly larger) for blob2 than for blob1, maybe just multiply by 1.01 and 0.99 ?

agnpy/time_evolution/tests/test_blob_expansion.py
```diff
@@ -0,0 +1,210 @@
...
+        blob2 = Blob(r_0, n_e=PowerLaw())
+        TimeEvolution(blob2, time, expansion=BlobExpansion(v_exp), method="heun",
+                      step_duration=2 * u.s).evaluate()
+        heun_reversed_analytically = blob2.gamma_e * (r_t / r_0).to_value("")
+
+        errors_euler = np.abs((euler_reversed_analytically - initial_gamma) / initial_gamma)
+        errors_heun = np.abs((heun_reversed_analytically - initial_gamma) / initial_gamma)
+        print("Average error Euler method", np.average(errors_euler))
+        print("Average error Heun  method", np.average(errors_heun))
+        assert np.average(errors_heun) < np.average(errors_euler)
```
jsitarek: I think it would be also worth to check if density is computed more accurately with Heun method, e.g. by comparing if the difference with the total number of electrons at the beginning is smaller.

agnpy/time_evolution/tests/test_blob_expansion.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def test_synchrotron_cooling_is_weaker_when_magnetic_field_decays(self):
```
jsitarek: not sure if this test is really needed

jsitarek (commented):
looks good, I left a few minor comments that we can discuss tomorrow

docs/time_evolution.rst
grzegorzbor: fixed

agnpy/time_evolution/tests/test_blob_expansion.py
grzegorzbor: Fixed

jsitarek (approved):

PR #202, opened 2026-08-03
Added blob light-travel-time integrator for a constant blob radius

grzegorzbor:


agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,355 @@
...
+    tau, W = window.kernel
+    return np.trapz(W.to_value("1/s"), tau.to_value("s"))
+
+
+class TestKernel:
+    """The geometric kernel: shape, limits and normalisation."""
+
+    def test_shape_and_normalisation(self):
+        R = 1e16 * u.cm
+        integrator = ltt_integrator_constant_radius(R, [1e15] * u.Hz, kernel_points_size=2001)
```
jsitarek: the number of kernel points is much larger than what one normally would use, and this probably makes the test a bit slow, how about decreasing it to 21 or 51 (and allowing a bit worse precision for integralif needed)?

agnpy/utils/conversion.py
```diff
@@ -34,6 +34,12 @@ def nu_to_epsilon_prime(nu, z=0, delta_D=1, m = m_e):
     epsilon = nu.to("", equivalencies=epsilon_eq)
     return (1 + z) * epsilon / delta_D
 
+def lab_time_to_blob_time(t_lab: u.Quantity, z=0, delta_D=1) -> u.Quantity:
```
jsitarek: maybe better to move this function to Blob class? you have z and delta_D already there so it will simplify the API and you either way need a Blob object to compute the SEDs. 
EDIT: especially that Ugo will need it for LIV part

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+---------------------------
+The quadrature converges at second order, and in practice the error is dominated by the kernel
+sampling rather than by the density of blob snapshots. Measured against a 6400-point reference
+for a synchrotron-cooling blob, with snapshot spacing between 1/100 and 1/15 of the window
+width (the two agreed to within a factor 1.3):
+
+    50 points -> 1e-3      200 points -> 8e-5      1000 points -> 2e-6
+    100 points -> 3e-4     500 points -> 1e-5      2000 points -> 6e-7
+
+The default of 500 therefore gives about 1e-5 relative accuracy, and raising it is cheap since
```
jsitarek: computing kernel is also not very complicated, because this is a simple analytical function. 
But later on to use this kernel you still need to average the SED over all those points and do it for all the frequencies (and possibly for multiple SEDs). The excessive accuracy here is likely not needed, because there are other simplifications in the model and e.g. we do not take into account that the radiation seen by electrons is not instantonously propagated in the whole blob.
So I think 50 points would be already more than needed and is a more reasonable default value

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+It means that, in order to calculate the smeared SED at a given time t, you must have a time margin [-tau, +tau]
+of time-evaluated blob data. For this reason, the object returned by `BlobLTTIntegrator.for_time` has additional
+`start_time` and `end_time` properties, which represent central_time +/- tau,
+which can help the caller know how far to run the simulation and which snapshots may be dropped:
+
+    integrator = ltt_integrator_constant_radius(blob.R_b, nu_obs=nu)
+
+    # 1. Suppose we want the SED at t=0
+    start_window = integrator.for_time(0 * u.s)
+    lead = -start_window.start_time # will be negetive, hence multiply by -1
```
jsitarek: I do not understand why you multiply by -1. If the central time is 0, then e.g. the integration time will be from -10 to +10, but then you need to keep this -10 as the starting point and evolve the blob for 10 - (-10) = 20 time units. I think you do it this way, just with an inverted definition of time, but it makes it more messy when you want to simulate a few periods one after another.

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+    0 + tau < sed_blob_time < total_blob_simulation_time - tau
+
+It means that, in order to calculate the smeared SED at a given time t, you must have a time margin [-tau, +tau]
+of time-evaluated blob data. For this reason, the object returned by `BlobLTTIntegrator.for_time` has additional
+`start_time` and `end_time` properties, which represent central_time +/- tau,
+which can help the caller know how far to run the simulation and which snapshots may be dropped:
+
+    integrator = ltt_integrator_constant_radius(blob.R_b, nu_obs=nu)
+
+    # 1. Suppose we want the SED at t=0
```
jsitarek: this t=0 is in blob frame, which should be repeated here. 
actually for 0 the observer clock gives the same value, making this example a bit confusing at point 3.

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+
+    # 3. Now again if you want to find SED at further time, repeat the same procedure, but remember about `lead` shift:
+    end_window = integrator.for_time(lead + end_time)
+    # run simulation until end_window.end_time, caching all blobs generated in the [end_window.start_time, end_window.end_time] range
+    sed = end_window.calc_sed(snapshots, snap_times)
+
+
+Warning:
+
+    The times passed to :meth:`BlobLTTWindow.calc_sed` must be on the same clock as the time
```
jsitarek: not only on the same clock, but it must be the clock in the blob frame

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+    def central_time(self) -> u.Quantity:
+        """Blob-frame time this SED is for; equal to the time passed to ``for_time``."""
+        return self._t0_s * u.s
+
+    @property
+    def start_time(self) -> u.Quantity:
+        """
+        Earliest blob-frame time contributing, emitted by the far side of the blob: t0 - R/c.
+
+        May be negative -- ``for_time(0)`` is the intended way to discover how much blob state
```
jsitarek, line 150: t_0 and end_time can be also negative, so I'm not sure if this sentence is explaining more or adding confusion.

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+        integrator = self._integrator
+        t_s = np.atleast_1d(times.to("s").value).astype(float)
+
+        if len(blobs) != t_s.size:
+            raise ValueError(
+                f"blobs and times must have the same length, got {len(blobs)} and {t_s.size}"
+            )
+        if t_s.size == 0:
+            raise ValueError("at least one blob snapshot is required")
+        if t_s.size > 1 and not np.all(np.diff(t_s) > 0):
```
jsitarek: check on t_s.size is not needed because it can never be negative and if it is 0 the previous exception is raised
EDIT: if t_s.size is 1 then the np.diff output is empty and np.all still returns true, but fine, it is more clear the way it is written now

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+            raise ValueError("times must be strictly increasing")
+
+        integrator._validate_radii(blobs, t_s)
+        self._validate_coverage(t_s, no_state_before_first)
+
+        table = integrator._sed_table(blobs, t_s)  # (n_nu, n_t)
+
+        tau_s, W_cgs = self._tau_s, self._W_cgs
```
jsitarek: why the copy is made and you do not simply use _tau_s and _W_cgs?

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+    The span of blob states needed to compute one light-travel-time smeared SED.
+
+    All times are BLOB-FRAME times, on the same clock as the time passed to
+    :meth:`BlobLTTIntegrator.for_time`. Obtained from that method rather than constructed
+    directly.
+    """
+
+    _integrator: "BlobLTTIntegrator"
+    _t0_s: float
```
jsitarek: I think there is no place where you need in this class pure, _tau_s - you always add _t0 to it (only once in the integration later on you specify _tau_s as an argument, but this does not matter because only np.diff of this is needed for integration, and this will be the same shifted or not by t0.
If you pass to this funtion directly t0_s + tau_s it will simplify slightly the code in multiple places.

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+        self._validate_coverage(t_s, no_state_before_first)
+
+        table = integrator._sed_table(blobs, t_s)  # (n_nu, n_t)
+
+        tau_s, W_cgs = self._tau_s, self._W_cgs
+        t_sample = self._t0_s + tau_s
+
+        if t_s.size == 1:
+            # Frozen state across the whole window; no interpolation needed.
+            weights = W_cgs
```
jsitarek: here you make a second copy of weights. if you do not want to modify the original precalculated ones only one of the two copies is sufficient.

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+            if no_state_before_first:
+                weights = np.where(t_sample < t_s[0], 0.0, W_cgs)
+            return (table[:, 0] * np.trapz(weights, tau_s)) << _SED_UNIT
+
+        if no_state_before_first:
+            # Clip only the top; interp1d zero-fills below times[0].
+            t_sample = np.minimum(t_sample, t_s[-1])
```
jsitarek: this one and L267 clip the times to max/min values, but the length of the array stays the same and SED will be evaluated multiple times for the same value. It might be faster to cut the array to only the needed range

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+
+class BlobLTTIntegrator:
+    """
+    Light-travel-time integrator for a spherical blob of constant radius.
+
+    Construct with :func:`ltt_integrator_constant_radius` rather than directly.
+
+    SEDs of individual snapshots are cached, so repeated ``for_time(...).calc_sed(...)`` calls
+    over an overlapping set of snapshots evaluate each snapshot only once. Cache entries for
+    snapshots no longer passed in are discarded, so the cache never keeps blobs alive.
```
jsitarek: "no longer passed" - you probably mean "not in the time range needed for integration over blob size"
"so the cache never keeps blobs alive" - I do not know what do you mean here

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+
+    def __init__(self, R: u.Quantity, nu_obs: u.Quantity, n_points: int, sed_flux_fn=None):
+        R_cm = float(R.to("cm").value)
```
jsitarek: why the cast to float? it should be already like this after .value

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+
+        This is a purely geometric query: it never inspects snapshots and never fails on
+        coverage grounds. ``for_time(0)`` legitimately returns a negative
+        :attr:`~BlobLTTWindow.start_time`, which is how you discover how much blob state is
+        needed before the nominal start of a run.
+        """
+        t0_s = float(t_blob.to("s").value)
+        return BlobLTTWindow(self, t0_s, self._tau_s, self._W_cgs)
+
+    def _validate_radii(self, blobs: Sequence, t_s: np.ndarray) -> None:
```
jsitarek: if you initialize the class with the first blob, you could read the radius directly out of it and would not have to check it here. The problem is just detecting expansion because now it is done in a separate class and original blob object does not know if it is expanding or not.

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+                )
+
+    def _sed_table(self, blobs: Sequence, t_s: np.ndarray) -> np.ndarray:
+        """
+        SED of every snapshot as an (n_nu, n_t) array, reusing cached rows.
+
+        A cached row is reused only when the blob at that time is the same object, so a
+        recomputed or replaced snapshot is re-evaluated. Afterwards the cache is pruned to the
+        times passed in, which both keeps it bounded and guarantees every cached id belongs to
+        a blob the caller still holds -- so ids cannot have been recycled behind our back.
```
jsitarek: I am wondering it this caching will not have some side effects in fitting. In that case you have all the time the same blob object, so the same id, but the values inside are modified by the minimalization function
Probably it will still work, because the blobs are deepcopied before being added.

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,424 @@
...
+                    self._sed_flux_fn(blob, self._nu_hz).to_value(_SED_UNIT), dtype=float
+                )
+            table[:, i] = row
+            fresh[key] = (id(blob), row)
+
+        self._sed_cache = fresh
+        return table
+
+
+def ltt_integrator_constant_radius(
```
jsitarek: why this function? can't you simply put all this in the constructor?

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,355 @@
...
+    def test_window_is_symmetric_about_the_central_time(self):
+        R, t0 = 1e16 * u.cm, 5e5 * u.s
+        integrator = ltt_integrator_constant_radius(R, [1e15] * u.Hz)
+        window = integrator.for_time(t0)
+
+        assert u.isclose(window.central_time, t0)
+        assert u.isclose(window.end_time - window.central_time, (R / c).to("s"), rtol=1e-12)
+        assert u.isclose(window.central_time - window.start_time, (R / c).to("s"), rtol=1e-12)
+
+    def test_for_time_zero_reports_the_required_lead(self):
```
jsitarek: if the previous test works, I do not think this one can fail

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,355 @@
...
+        lead = -integrator.for_time(0 * u.s).start_time
+        assert lead > 0
+        assert u.isclose(lead, (R / c).to("s"), rtol=1e-12)
+
+
+class TestCalcSed:
+    """Integration of blob states over the window."""
+
+    def test_constant_state_reproduces_the_plain_sed(self):
+        """With a frozen blob the smearing must be a no-op: the kernel integrates to 1."""
```
jsitarek: the kernel always integrates to 1, here what we integrate with the kernel is constant

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,355 @@
...
+
+        t0 = 2 * (R / c).to("s")
+        window = integrator.for_time(t0)
+        times = np.linspace(0, 4 * (R / c).to_value("s"), 4) * u.s
+        blobs = [deepcopy(blob) for _ in times]
+
+        sed = window.calc_sed(blobs, times)
+        expected = Synchrotron(blob).sed_flux(nu) + SynchrotronSelfCompton(blob).sed_flux(nu)
+
+        assert sed.unit.is_equivalent(SED_UNIT)
```
jsitarek: if this one fails, the next assert would fail as well, so you can skip the first one

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # fully before the step, halfway through it, fully after.
+        # The tolerance on the plateau is set by the kernel quadrature error (~4e-6 at the
+        # default 500 points), not by the smearing itself.
+        assert np.isclose(flux(3.0), 0.0, atol=1e-6)
+        assert np.isclose(flux(5.0), 0.5, rtol=2e-2)
+        assert np.isclose(flux(7.0), 1.0, rtol=1e-4)
+        # monotonically rising across the transition
+        rising = [flux(x) for x in np.linspace(4.0, 6.0, 9)]
+        assert all(a <= b + 1e-12 for a, b in zip(rising, rising[1:]))
```
jsitarek: you can write it simpler, something like assert np.all(np.diff(rising)>-1.e-12)

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,355 @@
...
+        window, blobs, times = self._setup(2e16 * u.cm)
+        with pytest.raises(ValueError, match="integrator was built for"):
+            window.calc_sed(blobs, times)
+
+    def test_calc_sed_does_not_modify_the_snapshots(self):
+        window, blobs, times = self._setup(1e16 * u.cm)
+        before = [blob.R_b.copy() for blob in blobs]
+        window.calc_sed(blobs, times)
+        for blob, original in zip(blobs, before):
+            assert u.isclose(blob.R_b, original, rtol=0, atol=0 * u.cm)
```
jsitarek: this test is strange, you check only R_b, and either way the value is the same in the blob generation and in the ltt class, so it would be very strange that it gets modified.

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,355 @@
...
+    def test_non_positive_radius(self):
+        with pytest.raises(ValueError, match="positive finite length"):
+            ltt_integrator_constant_radius(0 * u.cm, [1e15] * u.Hz)
+
+    def test_too_few_kernel_points(self):
+        with pytest.raises(ValueError, match="at least 2"):
+            ltt_integrator_constant_radius(1e16 * u.cm, [1e15] * u.Hz, kernel_points_size=1)
+
+
+class TestSingleSnapshot:
```
jsitarek: all this is a very strange situation. if you have only one blob at 't', and you must set no_state_before_first it means that <t you have 0, after t you do now know, and only in the precise moment of 't' you have an emission, so since it is infinitesimal time the integral should end up as 0. I would simply remove such a case of a single blob

jsitarek (commented):
in general looks fine, but some of the things seems to be done in a too complicated way, and the code is also trying to work with non-realistic cases, so I think it can be simplify somewhat - details in the specific comments

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: True. I changed it to 41 points, and reduced precision from 1e-5 to 1e-3

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: Changed to 50 as default, and removed this piece of documentation.

agnpy/utils/conversion.py
grzegorzbor: Good idea, Blob is much better place for it.

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: The idea was that we can have a set of functions:
ltt_integrator_constant_radius(R)
ltt_integrator_linear_expansion(R_0, expansion_rate)
ltt_integrator_variable_radius(R_fn) <- arbitrary R(t) function, needs numerical resolution

Factory methods are nicer than constructor in such case, because we can give them descriptive names.

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: Yes, it is only used in the kernel() property, which is only used for diagnostics in the test. I think i can remove it.

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: Here it's more about using the same "0 time" reference point, but indeed it's confusing, I reworded it.

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: Yes, the results is the same, it's just the difference how you look at it.
The main point is that "lead" in this example is an additional (absolute) time you need to include in the simultion, so the simultion must be longer. Because in point 2 I say "you must run it for lead+tail time". So if didn't took absolute value, i would need to say "-lead+tail" here, which would be probably more confusing.

So it's just different naming, the math is the same :)

I added additional comment to explain that i mean the "Absolute additional time needed for running the simulation"

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: ok, I added an explanation that all times are in blob frame

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: But passing negative t_0 is a weird case, I think we might even add a validation rule to reject it.
On contrary, negative start time may be received not only for the t_0 = 0 case, but also for some short steps, if central time is > 0 but still small compared to R/c, the start time can be < 0, so it must be handled correctly in the code, that's why I think it should be documented here.

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: it's not a copy, just a cheap local refrence alias, not to repeat "self." in arithemetic formulas below

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: No copy, just a reference, so it's cheap.
But it's not strictly needed, so i removed it, and reassigned W_cgs instead

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: yes it's better, i changed it

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: The integrator made with "ltt_integrator_constant_radius" method does not care much about the blob object, it only cares about the readius, which should be constant. That's why the factory method takes the radius, not blob, because radius is all it needs.

On the other hand, the sed interpolator needs a list of blobs, at different times. And these blobs must be deep copies.
So here is just an additional protection layer  which makes sure the blobs passed to the interpolator, really have the same radii as the radius expected by the integrator.

agnpy/time_evolution/blob_ltt_integration.py
jsitarek: It is not enough that you need to run it for a specific duration, but rather cover a specific time. With the current definition you need to run it from -lead to +tail. In what I proposed simply from lead to tail.

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: The blobs passed to the interpolator are always deep copies of the original blob, so it is safe, because minimization function cannot modify them

agnpy/time_evolution/blob_ltt_integration.py
jsitarek: yes, but it adds some complication to the code, and either way with the current code you would need either to modify BlobLTTIntegrator to add expansion in the constructor.

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: removed

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: removed

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: yes, comment corrected

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: yes, it would only make sense for expanding blob; removed

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: removed

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: removed

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: fixed

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: reworded

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: changed

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: I added "t0" to time evoltution as we agreed

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: Here the purpose of the clip is to align the first and last point of integration time points, with the first and last points of blob snapshot times list (we actually checked in earlier step they are equal, but with small numerical precision tolerance; but here even such small difference would cause interpolation to do extrapolation, or fill with zero, or throw; so we must get rid of any gap, and the only reason for clip is removing this tiny gap). 
I double checked, and also renamed variables to make them more obvious, and I don't see any possibility the SED would be evaluated multiple times.

docs/time_evolution.rst
```diff
@@ -148,13 +148,30 @@ indicate particle loss).
...
 The `method` parameter allows switching from the default Euler method to the more accurate, but slower, Heun
 method. The `distribution_change_callback` parameter accepts a user-defined function that is called after each
 time step and can be used to track the progress of the calculation.
 The `expansion` parameter enables modelling of an expanding blob - see the dedicated section below.
 
+The `t0` parameter sets the time on the blob's clock at which the simulation starts. It defaults to zero and does
+not affect the simulation itself; it only shifts the `blob_time` reported by the `TimeEvaluationResult` - both the
+one returned by `evaluate` and those passed to `distribution_change_callback`. For example, with a total duration
+of 100 s and `t0 = -30 s`, the reported times progress from just above -30 s up to 70 s. It must be a scalar
+`Quantity` whose unit is exactly the unit of `total_duration_time`, because the reported times are computed as
```
jsitarek: I think this limitation of the units is not needed (see the other comment)

agnpy/time_evolution/time_evolution.py
```diff
@@ -244,6 +255,18 @@ def __init__(self,
         if subgroups_initial_density is not None and initial_gamma_array is not None and subgroups_initial_density.shape[1] != len(initial_gamma_array):
             raise ValueError("Incorrect shape for subgroups_initial_density")
         self._distribution_change_callback = distribution_change_callback
+        # The reported times are t0 + elapsed, so astropy resolves them to t0's unit. Requiring
+        # t0 to share total_duration_time's unit keeps that unit predictable for the caller.
+        if t0 is None:
```
jsitarek: you do not need to make it so complicated. If you set the default to integer 0 astropy will not complain about adding 0 to time unit. And t0 is e.g. in days, and the other time is e.g. in seconds the astropy unit conversion will also work fine. You might check that somebody is not passing something strange if you check if t0.si.unit == "s"

agnpy/time_evolution/time_evolution.py
```diff
@@ -109,6 +109,16 @@ class TimeEvolution:
...
         You can use it, for example, for updating the distribution plot while the simulation is running.
+    t0 : Quantity
+        The time on blob's clock when the simulation starts. By default it is 0; but you may
+        change it to add some offset to the simulation time (positive or negative). It does not impact the simulation
+        process, but it will shift the time reported by the `TimeEvaluationResult` passed to the
+        `distribution_change_callback` function and returned at the end of the `evaluate` method.
+        For example, if total simulation duration is 100 s, and you set t0 = -30 s, then the
+        time reported by the `TimeEvaluationResult` at the end of the simulation will be 70 s, and intermediate times
+        reported in the callbacks will progress from -30s to 70s.
+        Must be a scalar Quantity whose unit is exactly `total_duration_time.unit`.
```
jsitarek: as explained below you do not need to match the units

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,378 @@
...
+    integration_window = integrator.for_time(blob.lab_time_to_blob_time(t_obs))
+    start = integration_window.start_time # (can be negative, but it's fine!)
+    end = integration_window.end_time
+    # if window start time > current simulation start time, catch up to it first:
+    TimeEvolution(blob, total_duration_time=(start-now), t0=now).evaluate()
+    # then make a list for blobs and corresponding times, and run the simulation from start till end
+    # of window. Seed the list with the state at `start`: the callback only fires after a step has
+    # been taken, so it never reports the state at t0 itself.
+    snapshots = [(start, deepcopy(blob))]
+    def callback(result):
```
jsitarek, line 46: in principle this code snippet could be also provided as a function - you use it a lot in various tests

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -0,0 +1,378 @@
...
+# the API boundary.
+_C_CGS = c_light.to("cm/s").value
+
+_SED_UNIT = u.Unit("erg / (cm2 s)")
+
+# Radii closer than this (relatively) are considered consistent when validating snapshots.
+_RADIUS_RTOL = 1e-6
+
+# Relative slack when checking that snapshots cover the window, to absorb float noise in times.
+_COVERAGE_RTOL = 1e-9
```
jsitarek: if it works it might be worth to use a smaller margin. with double precision 10^-14 should be fine, and if one passes to the function MJD as times (~60 000 days) and used time offsets of seconds this would end up as ~2 x 10^-10 fraction.

agnpy/time_evolution/tests/test_time_evolution.py
```diff
@@ -561,3 +561,70 @@ def test_total_electron_energy_loss_is_equal_to_total_radiation_energy(self):
...
+
+    def test_does_not_affect_the_simulation(self):
+        """Only the reported clock moves; the evolved distribution is identical."""
+        without = self._run(60 * u.s)
+        shifted = self._run(60 * u.s, t0=-1000 * u.s)
+        assert u.isclose(shifted.blob_time, -940 * u.s)
+        assert np.allclose(without.gamma, shifted.gamma)
+        assert u.allclose(without.density, shifted.density)
+
+    def test_reported_unit_follows_the_duration_unit(self):
```
jsitarek, line 613: if you modify this part about forcing the same units, the tests below can be removed/simplified

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,389 @@
...
+        return integrator.for_time(0 * u.s)
+
+    @pytest.mark.parametrize("n", [0, 1])
+    def test_fewer_than_two_snapshots(self, n):
+        """Interpolation needs two points; one cannot bracket a window anyway."""
+        snapshots = [(t * u.s, make_blob()) for t in range(n)]
+        with pytest.raises(ValueError, match="at least two"):
+            self._window().calc_sed(snapshots)
+
+    def test_non_monotonic_times(self):
```
jsitarek: if you want those two tests you can also write as a single with @parametrize and pass two lists of times as paramter

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -0,0 +1,389 @@
...
+                               max_energy_change_per_interval=0.2,
+                               distribution_change_callback=callback).evaluate()
+        if append_final:
+            blobs.append(deepcopy(blob))
+            times.append(result.blob_time)
+
+        times = u.Quantity(times)
+        return window, list(zip(times, blobs)), times
+
+    def test_workflow_produces_a_covered_window(self):
```
jsitarek, line 415: this and the next 2 tests are probably not needed. you test the whole workflow in the test above, and the covering of the window as well as excluding duplicated are checked in separate tests earlier on

agnpy/time_evolution/blob_ltt_integration.py
jsitarek: imagine that you are computing a big table of blobs with all the times because you want to compare SEDs at different times, so you have blobs with time 0, 5, 10, 15, 20, 25, 30, 35, 45. For the particular calculations of a single SED you need only times from 8 to 16 (and next you will use the same table for times e.g 29 to 37), 
If I get the code right, you table of times will look: 8, 8, 8, 15, 16, 16, 16, 16, 16. So you will compute t=8 3 times and t=16 5 times. Moreover it might actually change the times that are associated to the blobs.

jsitarek (commented):
second bunch of comments: the one about the clipping we should understand if we are on the same page - the others are just suggestions to consider

agnpy/time_evolution/time_evolution.py
grzegorzbor: Yes, 0 is not a problem, because units do not matter for 0.
But if you set the total_duration = 30s, and set t0 to a non-0 value, butwith different units, e.g. t0 = -1h then the total progress time reported in callbacks will be automatically converted to units of t0.
This will not break anything, will still work, but might be surprising for the user. So i thought it's better to add validation that if you specify t0, you should use the same units as total time, this way preventing surprise. 
But i can drop this check, and just add the note about the side effect of using two diffewrent time units.

jsitarek: but you return in the callback the time as Quantity, which means that it has the unit returned as well. So it does not matter if one returns 1 * u.min or 60* u.s, the returned object would still correspond to the same time. We would have a problem if you return just a float without unit, then indeed this would be very confusing and potentially problematic, or if somebody misuses Quantity (e.g. using .value instead of .to_value('s') ).

docs/time_evolution.rst
grzegorzbor: ok, removed

agnpy/time_evolution/time_evolution.py
grzegorzbor: fixed

agnpy/time_evolution/tests/test_time_evolution.py
grzegorzbor: done

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: removed

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: done

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: changed to 1e-14

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -370,3 +375,109 @@ def test_workflow_produces_a_covered_window(self):
...
+
+    def _lc(self):
+        """ Radius light-crossing time """
+        return (self.R / c).to_value("s")
+
+    def test_matches_the_manual_workflow(self):
+        """The automated loop must correctly proceed to start/end of window for each time point of interest:
+         0, 3, 4, 6 and 9 (in units or radius light-crossing), according to this timeline:
+
+              *           *   *       *           *
```
jsitarek, line 458: '6' and '9' are basically testing the same as '3', it would be interesting to put instead e.g. 4.5 to see if 4-4.5 is not calculated twice

agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -370,3 +375,109 @@ def test_workflow_produces_a_covered_window(self):
...
+            assert u.isclose(kwargs["total_duration_time"], expected_duration * lc * u.s)
+            if expected_t0 is not None:
+                assert u.isclose(kwargs.get("t0", 0 * u.s), expected_t0 * lc * u.s)
+            if expects_callback is not None:
+                callback = kwargs.get("distribution_change_callback")
+                assert (callback is not None) == expects_callback
+
+
+    def test_overlapping_windows_reuse_snapshots_instead_of_resimulating(self):
+        """Heavily overlapping windows must not each independently re-simulate from scratch."""
```
jsitarek: Heavily overlapping windows must not be independently re-simulated from scratch

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -376,3 +379,129 @@ def _sed_table(self, snapshots_s: Sequence[Tuple[float, Blob]]) -> np.ndarray:
...
+    times: u.Quantity,
+    nu_obs: u.Quantity,
+    *,
+    kernel_points_size: int = 50,
+    sed_flux_fn=None,
+    assume_steady_before_start: bool = True,
+    **time_evolution_kwargs,
+) -> list[u.Quantity]:
+    """
+   Utility function to run TimeEvolution over ``times`` and return the light-travel-time smeared SED at each one.
```
jsitarek: Utility function to run TimeEvolution starting from time = 0  over ``times`` ...

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -376,3 +379,129 @@ def _sed_table(self, snapshots_s: Sequence[Tuple[float, Blob]]) -> np.ndarray:
...
+    times : :class:`~astropy.units.Quantity`
+        Blob-frame times to compute the SED at, strictly increasing.
+    nu_obs : :class:`~astropy.units.Quantity`
+        Observed frequencies the SEDs are evaluated at; forwarded to :class:`BlobLTTIntegrator`.
+    kernel_points_size, sed_flux_fn
+        Forwarded to :class:`BlobLTTIntegrator`.
+    assume_steady_before_start : bool
+        The first requested time's window may reach before blob-frame time 0, if its
+        light-crossing margin is larger than ``times[0]`` itself. When ``True`` (the default),
+        ``blob``'s given state is treated as unchanged for however far back that window needs --
```
jsitarek: ...for as far back...

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -376,3 +379,129 @@ def _sed_table(self, snapshots_s: Sequence[Tuple[float, Blob]]) -> np.ndarray:
...
+            "them in time_evolution_kwargs"
+        )
+
+    times = u.Quantity(times)
+    if times.isscalar:
+        times = times.reshape(1)
+    if len(times) == 0:
+        return []
+    if len(times) > 1 and not np.all(np.diff(times.to_value("s")) > 0):
+        # We could sort them here, but it's probably better to let the user do it
```
jsitarek, line 444: ok, fine not to sort them, actually if we would like to also pass a different nu_obs for each point (this can happen, because the observations at different times can be at different energies as well), sorting would be quite error prone.

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: changed

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: changed

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: changed

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: done

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: 3 and 4 are overlapping windows;
4 and 6 are touching on the edges;
6 and 9 are disjoint.

So 4-4.5 seems the same as 3-4, both cover overlapping windows, right?

jsitarek (approved):

PR #203, opened 2026-08-24
BlobLTTIntegrator extended with added support for expanding blob radius

grzegorzbor:


agnpy/time_evolution/tests/test_blob_ltt_integration.py
```diff
@@ -86,6 +94,13 @@ def test_window_width_is_independent_of_time(self):
         ]
         assert u.allclose(u.Quantity(widths), 2 * (1e16 * u.cm / c).to("s"), rtol=1e-12)
 
+    def test_expanding_window_has_asymmetric_limits(self):
```
jsitarek, line 97: as we discussed today this test is very weak (it only checks sign) instead it is much better to include a test in which min and  max times are calculated by hand and compared with the output

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -5,27 +5,64 @@
...
-where F_std(nu, t') is the ordinary agnpy SED of a uniform blob in state t', and W is a purely
-geometric kernel: the cross-section of the sphere at offset tau, divided by the blob volume. For
-a blob of constant radius R, with V = 4/3 pi R^3,
+    ∫ W(ξ) * F_std(nu, ξ) dξ
 
-    W(tau) = pi c (R^2 - c^2 tau^2) / V = (3c / 4R) (1 - (c tau / R)^2)
+or, converting to the integration over dτ:
 
-The integration limits are such that tau spans [-R/c, +R/c].
+    F(nu, t_bc) = ∫ W(τ) * F_std(nu, t_bc + τ) dτ
```
jsitarek: dxi is not equal dtau, so W(xi) and W(tau) cannot be the same 'W' (there is a factor 'c' of difference.

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -5,27 +5,64 @@
...
 
+Linear expansion
+----------------
+Passing ``expansion`` to ``BlobLTTIntegrator`` models a blob whose radius grows
+at a constant rate BlobExpansion.v_exp, ``R(t) = R_0 + v_exp t``
+(it uses the same ``agnpy.time_evolution.BlobExpansion`` class as used by ``agnpy.time_evolution.TimeEvolution``).
+Writing: ``β_exp = v_exp/c``,``R_t = R(t_bc)`` and ``ρ = cτ/R_t`` (ρ is a line-of-sight depth measured in units of the R_t),
+we obtain:
+
+R(t_bc+τ) = R_t + v_exp·τ = R_t + β_exp·c·τ = R_t(1 + β_exp·ρ)
```
jsitarek: it is still the same formula as L45 just with beta instead of v. You could immediately introduce beta in L45 and avoid L50

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -5,27 +5,64 @@
...
 
 so a blob whose state does not change reproduces the ordinary SED.
 
 Because R and the sampling points are both fixed, the kernel is computed once when
 the integrator is built and reused for every requested time.
 
+Linear expansion
+----------------
+Passing ``expansion`` to ``BlobLTTIntegrator`` models a blob whose radius grows
+at a constant rate BlobExpansion.v_exp, ``R(t) = R_0 + v_exp t``
```
jsitarek: it is not clear if t and v_exp are measured in the frame of reference of the blob

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -5,27 +5,64 @@
 -------
 A blob does not radiate as a single snapshot. Photons that reach the observer together were
 emitted at different blob-frame times from different depths along the line of sight: the volume
-element at line-of-sight offset xi (positive = towards the observer) contributes emission from
-blob-frame time t0 + xi/c. Writing tau = xi/c, the observed SED at blob-frame time t0 is
+element at line-of-sight offset ξ (positive = towards the observer) contributes emission from
+blob-frame time t_bc + τ, where τ = ξ/c, and t_bc is the observed (lab) time transformed to the "blob-center" time
```
jsitarek, line 9: write already here that t_bc is the time corresponding to the observations from the center of the blob,

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -406,22 +486,25 @@ def calc_seds_over_time(
...
     assume_steady_before_start : bool
         The first requested time's window may reach before blob-frame time 0, if its
         light-crossing margin is larger than ``times[0]`` itself. When ``True`` (the default),
-        ``blob``'s given state is treated as unchanged for as far back that window needs --
+        ``blob``'s particle state is treated as unchanged for as far back that window needs --
         the same assumption the manual workflow above makes implicitly by seeding at
-        ``start_time`` rather than at 0. When ``False``, that situation raises instead, naming
-        how much earlier ``blob``'s history would need to start.
+        ``start_time`` rather than at 0. Its radius is not an assumption, though: under
+        ``expansion`` the backdated snapshot's ``R_b`` is set to ``integrator.radius_at(start)``,
```
jsitarek, line 511: I do not think you really need this special case. If the blob is expanding assuming that it has a steady state is hard to justify. You can just throw an error if somebody switches on this option with expanding blob

jsitarek (changes requested):
Ok, just some of the descriptions are to be clarified

agnpy/time_evolution/blob_ltt_integration.py
jsitarek: ok, but you still miss changing W to W' in this formula

agnpy/time_evolution/blob_ltt_integration.py
```diff
@@ -7,32 +7,41 @@
...
+W'(τ) = cA(τ)/V(τ)
+A(τ) = π(R^2 - ξ^2) = π(R^2 - (τc)^2)
+
+A note on time measuring:
+-------------------------
+Let us pick any reference time in the observer (lab) frame as t_obs = 0, which is also t_blob = 0 in the blob frame.
+Then we pick any other time in the lab frame and put the question: what is the observed SED at this time?
+The observed SED is a combination of SEDs emitted from different slices of blob at different times in the blob frame.
+Hence, we name ``t_bc`` a blob-frame time corresponding to the emission from the blob's slice containing blob's center,
+observed at the requested time in the lab frame. We **could** call it "t_0" as well, but this could be confused with the t=0 point,
```
jsitarek: the last two sentences (from "We could...) you can drop

grzegorzbor: done

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: done

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: corrected now, pls check

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: i reprhased it, should be clear now, can you verify?

agnpy/time_evolution/blob_ltt_integration.py
grzegorzbor: Yes this makes no physical sense. Removed.

agnpy/time_evolution/tests/test_blob_ltt_integration.py
grzegorzbor: corrected

